PR #63, opened 2020-07-03
Fix invalid Pip VCS requirement specs

### LINKED ISSUE #62: VCS lines in requirements_base.txt are invalid
Pip VCS requirement specifications require the URL to specify what the egg name is, otherwise the line is invalid. See https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support.

I originally ran into this while trying to package Amulet for [Flatpak](https://flatpak.org/) - [this utility script](https://github.com/flatpak/flatpak-builder-tools/blob/master/pip/flatpak-pip-generator) chokes on `requirements_base.txt` because all the Git dependencies have this problem, making them invalid. See flatpak/flatpak-builder-tools#134 and the associated PR for details.

To fix this, they all need to have `#egg=<pkgname>` appended to them. I will be sending patches to make this change shortly but I needed a bug number for the branch name.

strugee:
We add `#egg=<pkgname>` at the end of all these to make the specification valid. See https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support for details.

I have verified that `pip3 install -r requirements_base.txt` works OK after this change. Note also that this is based on PR #59 because otherwise there would be merge conflicts. If there's stuff that needs to be addressed in #59 though I'd be happy to rebase this patch on `master` so #59 can land on top of this PR instead of the other way around.

Fixes #62

### DIFF
--- a/requirements_base.txt
+++ b/requirements_base.txt
@@ -1,6 +1,6 @@
 numpy==1.17.4
 pyopengl==3.1.5
-git+git://github.com/Amulet-Team/Amulet-Core.git
-git+git://github.com/Amulet-Team/Amulet-NBT.git
-git+git://github.com/gentlegiantJGC/PyMCTranslate.git
-git+git://github.com/gentlegiantJGC/Minecraft-Model-Reader.git
+git+git://github.com/Amulet-Team/Amulet-Core.git@4986e49c143abded07e548b486d71c5085a77d41
+git+git://github.com/Amulet-Team/Amulet-NBT.git@138a71cb1245b07b9a9afd8dfca6ac73bd659a78
+git+git://github.com/gentlegiantJGC/PyMCTranslate.git@1d4601f85f994af230f979e419bd871ae68ba5c5
+git+git://github.com/gentlegiantJGC/Minecraft-Model-Reader.git@7b2a2fe8b04ff742c57fec41bfdd38bc97b1996e
--- a/requirements_base.txt
+++ b/requirements_base.txt
@@ -1,6 +1,6 @@
 numpy==1.17.4
 pyopengl==3.1.5
-git+git://github.com/Amulet-Team/Amulet-Core.git@4986e49c143abded07e548b486d71c5085a77d41
-git+git://github.com/Amulet-Team/Amulet-NBT.git@138a71cb1245b07b9a9afd8dfca6ac73bd659a78
-git+git://github.com/gentlegiantJGC/PyMCTranslate.git@1d4601f85f994af230f979e419bd871ae68ba5c5
-git+git://github.com/gentlegiantJGC/Minecraft-Model-Reader.git@7b2a2fe8b04ff742c57fec41bfdd38bc97b1996e
+git+git://github.com/Amulet-Team/Amulet-Core.git@4986e49c143abded07e548b486d71c5085a77d41#egg=amulet-core
+git+git://github.com/Amulet-Team/Amulet-NBT.git@138a71cb1245b07b9a9afd8dfca6ac73bd659a78#egg=amulet-nbt
+git+git://github.com/gentlegiantJGC/PyMCTranslate.git@1d4601f85f994af230f979e419bd871ae68ba5c5#egg=pymctranslate
+git+git://github.com/gentlegiantJGC/Minecraft-Model-Reader.git@7b2a2fe8b04ff742c57fec41bfdd38bc97b1996e#egg=minecraft-model-reader


Podshot:
I was not aware of the `#egg=` issue, but thank you for bringing it our attention! I agree with your point that you made in #59 that our builds should be reproducible and deterministic and shouldn't rely on the current state at build time of any of our dependency repositories but on the exact commits for each. Our build system in it's current state is manually triggered by pushing to master and changing `/amulet_map_editor/version` which usually means the dependencies are up-to-date at the time of the build being ran, but does present potential issues if one or more are out-of-sync. 

I feel the `egg=` change is straightforward while pinning commit hashes may not be due to frequent changes in our dependencies and how different branches of each repository link to other branches until everything is merged to the respective `master`/`develop` branches. We could have a script do this, but that again brings up issues with branches, etc.

Overall I think this PR is a very much needed enhancement, but it will be a complicated issue to resolve overall. @gentlegiantJGC I would your input on this since I know you've done a lot of work spanning the editor, core, and your pymctranslate repositories, so I feel this will affect your workflow.

strugee:
@Podshot thanks for the response :)

One thing we could do as a compromise is to pin git commits only for releases. That won't make `git bisect` work as well as it could, but it solves the reproducability problem and allows manual bisections at least for versioned releases. Also I've realized that if we really want to make this work we will also need to [`pip freeze`](https://pip.pypa.io/en/stable/reference/pip_freeze/) the whole dependency tree. That could also be done only at release time to avoid impacting workflows though.

Since this is complicated, I propose I redo this PR so its dependency on #59 is reversed while we sort that out/wait for @gentlegiantJGC to weigh in. That way while we figure out what the best way to do that is, this fix will have already gone in and won't be blocking my packaging work.

By the way, would you be interested in taking [Flatpak packaging](https://flatpak.org/) files upstream?

gentlegiantJGC:
To confirm there are two issues here. The `egg` issue is to specify the branch which is an easy fix since that will always be the master branch.

The other issue is if someone wants to install an older version it will get the newer version if it has been updated.
Is there a way we can create a script that will push each of the dev branches to master, get the commit id and update the requirements to include it?

gentlegiantJGC:
The pipeline for an updating script would be as follows:

user needs to manually merge branches into the dev branches for:
```
Minecraft-Model-Reader
Amulet-NBT
PyMCTranslate
Amulet-Core
Amulet-Map-Editor
```

1) Ask user for the new version number
2) Write that version number to Amulet-Map-Editor/amulet_map_editor/version
3) Commit that change to Minecraft-Model-Reader:dev if changed
4) Merge Minecraft-Model-Reader:dev into Minecraft-Model-Reader:master
5) Update PyMCTranslate:dev requirements.txt
	- Amulet-NBT:master
6) Merge PyMCTranslate:dev into PyMCTranslate:master
7) Update Amulet-Core:develop requirements.txt
	- Amulet-NBT:master
	- PyMCTranslate:master
8) Merge Amulet-Core:develop into Amulet-Core:master
9) Update Amulet-Map-Editor:dev requirements.txt
	- Amulet-Core:master
	- Amulet-NBT:master
	- PyMCTranslate:master
	- Minecraft-Model-Reader:master
10) Merge Amulet-Map-Editor:dev into Amulet-Map-Editor:master
11) Automatic build system should kick in

Podshot:
We could possibly have a script that could perform that workflow.

@strugee I would be interested in taking a look at that, anything to improve our support on Linux would be appreciated. I would like to keep the building process on our TravisCI page if it's possible, but I assume that's a pretty easy requirement to reach. Also, we're currently in the process of re-licensing the project to a new license, so before we merge any of your PRs do we have your permission to merge your contributions and then re-license them to a new license at some point in the future? We haven't determined which license we are switching to yet, but we can let you know which one we will re-license to when we make the decision

strugee:
> The `egg` issue is to specify the branch which is an easy fix since that will always be the master branch.

No, this isn't right. `#egg=<pkgname>` on VCS requirements isn't about the git branch, it's about the name of the package, i.e. what's in `setup.py` in the dependency's repository.

As for the proposed pipeline I'm having a bit of trouble following it. Can you clarify the difference between the `dev` and `master` branches in your workflow? And am I understanding correctly that when the `version` file in Amulet-Map-Editor is bumped, that change needs to happen in deps too?

strugee:
> I would be interested in taking a look at that, anything to improve our support on Linux would be appreciated. I would like to keep the building process on our TravisCI page if it's possible, but I assume that's a pretty easy requirement to reach.

Yeah, we could do that. We'd have to give the signing keys to Travis though. I'd be willing to manage the releases onto [Flathub](https://flathub.org/home) if that would work for you, so we could avoid that. Otherwise putting it on Travis would work fine.

> Also, we're currently in the process of re-licensing the project to a new license, so before we merge any of your PRs do we have your permission to merge your contributions and then re-license them to a new license at some point in the future? We haven't determined which license we are switching to yet, but we can let you know which one we will re-license to when we make the decision

I would be fine with any license on the Free Software Foundation's [list of free software licenses](https://www.gnu.org/licenses/license-list.html#SoftwareLicenses). That being said I would prefer a copyleft license, although of course it's your project so it's up to you :)

gentlegiantJGC:
The license is still undecided but the plan has always been to have it free for use by the general public in non-commercial cases and to have the source code publicly available. Those parts will not change but we want to retain some ownership of our work and have the ability to add some other restrictions.

To have your pull requests considered you will need to acknowledge you understand that and agree to it being relicensed under any licence.

strugee:
> To have your pull requests considered you will need to acknowledge you understand that and agree to it being relicensed under any licence.

Any contribution I make is made available under the terms of the current license, the MIT license. The MIT license allows sublicensing subject to the terms of the attribution clause of the license.

I'm really sorry to have to be so pedantic, but [users' freedom and rights](https://www.gnu.org/philosophy/free-sw.html), the [power relationships](https://www.gnu.org/philosophy/free-software-even-more-important.html) that exist in software production, and the idea of a shared commons are all important. That is why I cannot waive any more requirements beyond what the MIT license - which is already _very_ permissive - allows. I feel that even if you might not agree with the freedom-related arguments I'm making, the attribution requirement is quite reasonable.

If you _do_ insist on proprietarizing this project, please consider using something like the [Business Source License](https://mariadb.com/bsl11/). This license can be used to permit most uses except commercial ones, and the commercial restrictions terminate at a set date such that four years (or sooner if you so choose) after the release of a particular version of the software, that release becomes freedom-respecting software. This means that you can set the terms you want for a full four years after a version of the software is released, and then it is released to the software commons. Using a well-known license like the BSL also helps alleviate the problems associated with [license proliferation](https://en.wikipedia.org/wiki/License_proliferation).

gentlegiantJGC:
This has been fixed now with proper versioning and PyPi support

PR #64, opened 2020-07-03
Tell people to run `python3` in the README

strugee:
On many Linux systems, `/usr/bin/python` still points to Python 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,9 @@ Extract the contained folder to a location on your computer and run the executab
 
 In order to run Amulet from source, you will need to install the following packages and the dependancies for those packages:
 
-The following can be easily installed using `python -m pip install -r requirements.txt`
+The following can be easily installed using `python3 -m pip install -r requirements.txt`
 
-For linux run `python -m pip install -r requirements_linux.txt`
+For linux run `python3 -m pip install -r requirements_linux.txt`
 
 - numpy
 - wxpython
@@ -33,7 +33,7 @@ For linux run `python -m pip install -r requirements_linux.txt`
 - [PyMCTranslate](https://github.com/gentlegiantJGC/PyMCTranslate)  The library to handle block, block entity, entity and biome translation. between versions
 - [Minecraft-Model-Reader](https://github.com/gentlegiantJGC/Minecraft-Model-Reader)  The library to handle loading block models and textures from a resource pack for use in the renderer.
 
-run `python main.py`
+run `python3 main.py`
 
 ## Contributing
 


gentlegiantJGC:
Does the same apply to windows? I think I have manually added python to my path as just `python` so I don't know what the python 3 installer normally adds to the path.

strugee:
Good question, I don't really know. If you have both installed I think it just depends on what order they're in in `$PATH`, which probably was determined by what order you installed 2 and 3 in, assuming you checked the box to add Python to `$PATH` in the installer - I can't remember if this is on by default.

As long as the Python 3 installer adds a `python3` binary it should work ok no matter what the order is. I don't have a Windows machine to test on though, unfortunately.

Ex-Ark:
If you're wondering about windows I did some tests on windows 10 with no previously installed python :

The python 2 installer does not set the PATH by default
image

---

Same thing for python 3

image

If the user (like I did) chose to add python to PATH every time then here's the result (first python 2 installed then python3):

image

Even if python 3 comes first in the PATH the old python is still called : 

image

Also there is no `python3` binary executable, only `python`

Here are all exe in my python3 directory added to PATH:
```rb
# Python38\Scripts
easy_install.exe
easy_install-3.8.exe
pip.exe
pip3.8.exe
pip3.exe

# Python38\
python.exe
pythonw.exe
```

strugee:
Hm. Wow, ok. That's disappointing. Thanks for testing :)

Not really sure what to do here. I guess we're going to have to tell people on Windows to do something different? At any rate the current instructions are also broken for them if their `python` binary is Python 2. I _think_ with https://github.com/Amulet-Team/Amulet-Map-Editor/pull/66/commits/87e008ca4032100543b95508470f5082eada020d applied, double-clicking `main.py` on Windows should Just Work™?

> Even if python 3 comes first in the PATH the old python is still called :

Weird. That doesn't make any sense to me. Did you log out and log in again? :thinking:

Ex-Ark:
@strugee 

Yes , confirmed again today, after a clean reboot, the default python version for me is still `2`

There's this [really good SO answer](https://stackoverflow.com/a/5088548/6324558) that suggest using a shebang at the beginning of the `.py` file to force the correct python version.

Curiously, reading the answer I tried `python --version` that gives me `2.7.18`, and `py --version` that gives me `3.8.3`
I think the python version will always be different based on the user setup, OS, PATH and so on ..

If I have some time later I'll try putting a shebang at the start of amulet entrypoint, we'll see how my python version behave.

gentlegiantJGC:
The best answer is to make a section per is and describe the native way. If a user has it set up differently they should know what they are doing

strugee:
> There's this really good SO answer that suggest using a shebang at the beginning of the .py file to force the correct python version.

Yep, that links to the same documentation I was looking at a few days ago. The docs also say that `py` was added in Python 3.3 which is why `py --version` is reporting Python 3.8 for you: that utility doesn't exist in the 2.7 distribution. https://github.com/Amulet-Team/Amulet-Map-Editor/commit/87e008ca4032100543b95508470f5082eada020d does exactly what you said - it adds a shebang line. That is why I believe it will make double-clicks Just Work under Windows.

I'm still curious why you're getting Python 2 as `python` though. Can you run `echo %PATH%` in `cmd` and paste the results? (Or if you prefer PowerShell: `$Env:Path`)

strugee:
> The best answer is to make a section per OS and describe the native way. If a user has it set up differently they should know what they are doing

I mean, what's the "native way" on Windows though? I guess double-clicking and letting the `.py` file association work?

Ex-Ark:
> Can you run echo %PATH% in cmd and paste the results?

It seems my `%PATH%` is indeed incorrect.
It lists python 2.7 first, but that's not the order I have given in the windows env path declaration. (top to bottom : python3 , python2)

Anyway the shebang looks like the perfect solution
Here's what I tried in local (windows 10, I have cygwin installed but I used regular `cmd` to test the program)

```py
#!/usr/bin/env python3

import sys
if sys.version_info[0] < 3:
    raise Exception("Must be using Python 3")
```

It works on my windows machine, python 3 is used for the program and I am not seeing any error

If I do instead :
```py
#!/usr/bin/env python2

import sys
if sys.version_info[0] < 3:
    raise Exception("Must be using Python 3")

```

then it raises 

```

  File "C:\monkey\strugee_editor\main.py", line 5, in <module>
    raise Exception("Must be using Python 3")
Exception: Must be using Python 3
```

In short the shebang works on windows as far as I can tell

Ex-Ark:
Apparently the shebang supports comes from the  py launcher (when associated on win with `.py` files)
It adds virtual commands  : 
- /usr/bin/env python
 - /usr/bin/python
 - /usr/local/bin/python
  - python

documentation : https://docs.python.org/3/using/windows.html#shebang-lines

So it should be cross platform compatible

strugee:
Thanks for testing. I will amend the patch in the morning.

gentlegiantJGC:
I have added the suggested shebang and sys logic in 42757cd26ffed215491dd3c9126c2c93f5da326f
I have updated the readme with the suggested changes 12ada1a30d4f311306ea96c462f29b049c802354

PR #66, opened 2020-07-04
Make this a Pip-installable package (+ other misc fixes)

strugee:
The main goal here is to make `pip3 install .` work properly. I need this because it makes Flatpak packaging for Linux a whole lot easier to manage. I've broken each logical change into an individual commit - each commit has pretty good explanations for why the change is needed.

The high-level summary of this patch series is:

* Misc. tweaks, some (the `__init__.py` changes) to get ready for `setup.py` and some (`chmod +x main.py`) because it's just probably a good idea
* Introduce `setup.py` so that `pip install .` works
* Make `pip install .` install a wrapper script in $PREFIX/bin to run the app, and move the code in `main.py` around to make that happen
* Add `LICENSE`, which is unrelated but I didn't feel like bothering with a whole separate PR (sorry)

Draft status because I have yet to conduct a lot of detailed testing on this PR and I need to go through the diff and add some comments to point out questions I had while preparing these patches. Also I need to run `black`.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018-2020 Amulet Team
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/amulet_map_editor/__init__.py
+++ b/amulet_map_editor/__init__.py
@@ -3,6 +3,16 @@ from .util.log import log
 from . import lang
 import os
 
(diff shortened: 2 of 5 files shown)


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018-2020 Amulet Team
```
strugee, line 3: Is the date here correct? I guessed

amulet_map_editor/__init__.py
```diff
@@ -3,6 +3,16 @@
...
 
+__description__ = 'A new Minecraft world editor and converter that supports all versions since Java 1.12 and Bedrock 1.7'
+__url__ = "https://www.amulet-editor.com/"
+__uri__ = __url__
+__doc__ = __description__ + " <" + __uri__ + ">"
+
+__author__ = 'Amulet Team'
+
+__license__ = "MIT"
+__copyright__ = "Copyright (c) 2018-2020 Amulet Team"
```
strugee, line 14: Same date question here

setup.py
```diff
@@ -0,0 +1,95 @@
...
+KEYWORDS = ["minecraft", "map editor"]
+CLASSIFIERS = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: End Users/Desktop",
+    "Natural Language :: English",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: Microsoft :: Windows",
+    "Operating System :: MacOS",
+    "Operating System :: MacOS :: MacOS X",
+    "Operating System :: POSIX :: Linux",
```
strugee, line 22: I listed each of these individually because that's what Travis tests on. I'm not sure if I should just instead say it's cross-platform? Is there any platform-specific code? (Not that this _super_ matters since it's not going on PyPI anyway)

setup.py
```diff
@@ -0,0 +1,95 @@
...
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Games/Entertainment",
+    "Private :: Do Not Upload",
```
strugee, line 30: Note: this will prevent accidental uploads to PyPI which I assume is the desired behaviour, at least until the team decides otherwise.

setup.py
```diff
@@ -0,0 +1,95 @@
...
+    setup(
+        name=NAME,
+        description=find_meta("description"),
+        license=find_meta("license"),
+        url=find_meta("url"),
+        version=read(os.path.join("amulet_map_editor", "version")),
+        author=find_meta("author"),
+#        author_email=find_meta("email"),
+        maintainer=find_meta("author"),
+#        maintainer_email=find_meta("email"),
```
strugee, line 79: I left these out because I didn't know what email to put here. If there's an obvious choice though I'd love to fix it

gentlegiantJGC:
Fixed

PR #76, opened 2020-08-04
New Pick Block UI and better image support

gentlegiantJGC:


gentlegiantJGC:
Once this is sorted we should merge it and create a new release

Issue #99, opened 2020-09-04
Save UI resized window settings when user reopens Amulet windows


qmagnet:
The UI window opens fairly smally and cuts off the text for files. If a user drags the window to resize, next time the program opens, it would preferred if it remembered the window size.

**Window example of cut-off images and text**
![UI auto size](https://user-images.githubusercontent.com/13902631/92289073-386be680-eedd-11ea-9229-214160a1c2e8.png)

gentlegiantJGC:
This should be fixed now

PR #100, opened 2020-09-05
Preserve positions and sizes of windows between program launches

Podshot:
This system allows for any custom window/dialog that we have created to save/restore it's position/size when initialized without requiring a full rewrite of the target class or our base UI classes.

Solution for #99 

### Changes
- For any custom UI window/dialog, adding the `@preserve_ui_preferences` decorator (from the `amulet_map_editor.api.wx.util.ui_preferences` module) to the class declaration modifies the classes `__init__` method to preserve the previous behaviour while adding the additional attributes needed and 3 new event handlers (all contained within the `ui_preferences` module)
  - No other changes are needed to the target UI class to achieve this behaviour
- All data is stored the `config/window_preferences.config` file as a JSON object once Amulet exits
  - _Note: Unexpected program exits/crashes still need testing_
- When a decorated UI class is initialized, if it's entry exists in the configuration from the same launch instance or a previous one, it will have it's position and size changed to match, otherwise it's default values will be utilized and won't be stored until changed by the user

### DIFF
--- a/amulet_map_editor/api/framework/amulet_ui.py
+++ b/amulet_map_editor/api/framework/amulet_ui.py
@@ -10,6 +10,7 @@ from amulet_map_editor.api.framework.pages import WorldPageUI
 from .pages import AmuletMainMenu, BasePageUI
 
 from amulet_map_editor.api import image
+from amulet_map_editor.api.wx.util.ui_preferences import preserve_ui_preferences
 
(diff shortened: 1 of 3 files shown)


gentlegiantJGC:
This is cool. Can the config section use `api.config` rather than doing its own thing

PR #102, opened 2020-09-07
Deselect first switches to the select tool when other tools are enabled

gentlegiantJGC:
If any other tool is enabled (eg the operation tool) and the deselect action is run it will switch to the select tool.
Once in the select tool the command will deselect the actual box.
This is essentially just a shortcut to get to the select tool.


Podshot (approved):
Looks good to me! Merging it now!

Issue #46, opened 2020-05-28
Feature: Setbiome


Dy1anBy1an:
The feature to set an area that does not have to be chunk aligned to a specific biome, including all the new biomes introduced in 1,13+, The Void only on Java, and Legacy Frozen Ocean only on Bedrock would be a very nice feature to have.

A2va:
In the same idea it would be great to implement the features of BiomePainter (https://mblaine.github.io/BiomePainter/).

gentlegiantJGC:
That is the kind of thing I had in mind. Biomes in the newer Java versions are 3d so we would also need to account for that somehow

PR #103, opened 2020-09-07
Added "Set Biome" Operation

naor2013:


felix91gr:
Hi! I'm passing by.

So that these two are linked together: This solves #46

gentlegiantJGC:
I have had a bit of a play around and it it seems to work well. Need to do some more testing before merging.

One interesting thing I have discovered while messing with it is that in Java with 3D biomes the biome in the overworld is determined by the first layer of the array and not the whole array. This is not the case in the nether where the full array is used.

Basically it is stored as a 3D array everywhere but in the overworld (and presumably the end but I have not tested) only the first layer is used as a 2D array and the rest is ignored.

PR #143, opened 2020-12-09
Added auto-updater

Podshot:
This adds an external auto-updater that can be used when the user is notified that a new update has been released. Currently the updater only has support for Windows installations and has only been tested on Windows. The updater will download the latest update, backup the current installation, copy over plugins/operations/settings, then restart Amulet.

- [Amulet Updater codebase](https://github.com/Podshot/AmuletUpdater)
  - The updater is written in Java and the release comes with a bundled OpenJDK distribution
- [Update operation](https://gist.github.com/Podshot/f415300561d43d203158ce786d95e0a3)
  - An testing operation that allows for others to test of the update process on an already built past release, this is not included in this pull request

### DIFF
--- a/amulet_map_editor/api/framework/update_check.py
+++ b/amulet_map_editor/api/framework/update_check.py
@@ -11,6 +11,7 @@ import sys
 import zipfile
 import subprocess
 import glob
+import platform
 
 import wx
 
@@ -18,6 +19,9 @@ from amulet_map_editor.api.logging import log
 
 URL = "http://api.github.com/repos/Amulet-Team/Amulet-Map-Editor/releases"
 DOWNLOAD_URL = "https://github.com/Podshot/AmuletUpdater/releases/download/latest/AmuletUpdater.zip"
+
+NOT_RUNNING_FROM_SOURCE = getattr(sys, "frozen", False) and hasattr(sys, "_MEIPASS")
+AUTOUPDATER_SUPPORTED_OS = platform.system() in ("Windows",)
 
 VERSION_REGEX = re.compile(
     r"^v?(?P<major>\d+)\.(?P<minor>\d+)(\.(?P<patch>\d+))?(\.(?P<bugfix>\d+))?(b(?P<beta>\d+))?$"
@@ -133,13 +137,13 @@ class UpdateDialog(wx.Dialog):
         sizer_2 = wx.BoxSizer(wx.HORIZONTAL)
 
         update_page_button = wx.Button(self, label="Go to Download Page")
-        if not __debug__:
+        if NOT_RUNNING_FROM_SOURCE and AUTOUPDATER_SUPPORTED_OS:
             updater_button = wx.Button(self, label="Run Auto Updater")
         ok_button = wx.Button(self, label="Ok")
 
         sizer_2.Add(update_page_button, 0, wx.ALL, 5)
         sizer_2.Add((0, 0), 1, wx.EXPAND, 5)
-        if not __debug__:
+        if NOT_RUNNING_FROM_SOURCE and AUTOUPDATER_SUPPORTED_OS:
             sizer_2.Add(updater_button, 0, wx.ALL, 5)
             sizer_2.Add((0, 0), 1, wx.EXPAND, 5)
         sizer_2.Add(ok_button, 0, wx.ALL, 5)
@@ -153,7 +157,7 @@ class UpdateDialog(wx.Dialog):
         update_page_button.Bind(
             wx.EVT_BUTTON, lambda evt: self.goto_download_page(new_version, evt)
         )
-        if not __debug__:
+        if NOT_RUNNING_FROM_SOURCE and AUTOUPDATER_SUPPORTED_OS:
             updater_button.Bind(
                 wx.EVT_BUTTON,
                 lambda evt: self.auto_update(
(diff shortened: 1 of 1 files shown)


amulet_map_editor/api/framework/update_check.py
```diff
@@ -121,11 +130,16 @@ def __init__(self, parent, current_version: str, new_version: str):
 
         sizer_2 = wx.BoxSizer(wx.HORIZONTAL)
 
-        update_button = wx.Button(self, label="Go to Download Page")
+        update_page_button = wx.Button(self, label="Go to Download Page")
+        if not __debug__:
```
gentlegiantJGC, line 136: I believe `__debug__` is always True unless we define a flag when running python that we currently don't do

amulet_map_editor/api/framework/update_check.py
```diff
@@ -145,9 +163,50 @@ def goto_download_page(new_version, _):
...
         )
 
+    @staticmethod
+    def auto_update(current_version, target_version, is_beta, _):
+        working_directory = os.path.dirname(sys.executable)
+        temp_dir = os.path.join(working_directory, "updater-tmp")
+        try:
+            os.mkdir(temp_dir)
+        except FileExistsError:
+            pass
```
gentlegiantJGC, line 178: I generally prefer using `os.makedirs(temp_dir, exist_ok=True)`

amulet_map_editor/api/framework/update_check.py
```diff
@@ -6,10 +6,19 @@
...
+import zipfile
+import subprocess
+import glob
 
 import wx
 
+from amulet_map_editor.api.logging import log
+
 URL = "http://api.github.com/repos/Amulet-Team/Amulet-Map-Editor/releases"
+DOWNLOAD_URL = "https://github.com/Podshot/AmuletUpdater/releases/download/latest/AmuletUpdater.zip"
```
gentlegiantJGC, line 20: Can we migrate this so it is part of Amulet-Team?

amulet_map_editor/api/framework/update_check.py
```diff
@@ -145,9 +163,50 @@ def goto_download_page(new_version, _):
...
+    @staticmethod
+    def auto_update(current_version, target_version, is_beta, _):
+        working_directory = os.path.dirname(sys.executable)
+        temp_dir = os.path.join(working_directory, "updater-tmp")
+        try:
+            os.mkdir(temp_dir)
+        except FileExistsError:
+            pass
+
+        updater_zip = os.path.join(temp_dir, "AmuletUpdater.zip")
```
gentlegiantJGC, line 180: Since this is windows specific currently, can we add an OS check and run the old logic for the other OSs?

gentlegiantJGC:
I haven't tried running this but the code looks pretty good. You use `__debug__` quite a bit but that will be `True` all the time unless a specific argument is added which sets it to `False` and also disables asserts.

Is there any way we can cut down on the file size of the updater? 193MB is quite a bit especially with github's relatively slow download speeds.
When unpacked over 300MB is the JDK. Do we need JDK and not JRE or JVM (I don't know much of what I am doing here)
If not can we have them download separately and cache Java so it doesn't need downloading next time?

Podshot:
> Is there any way we can cut down on the file size of the updater? 193MB is quite a bit especially with github's relatively slow download speeds.
> When unpacked over 300MB is the JDK. Do we need JDK and not JRE or JVM (I don't know much of what I am doing here)
> If not can we have them download separately and cache Java so it doesn't need downloading next time?

I can remove the JDK from the release zip file to reduce it's size. As for using the JDK, due to the terms of the OpenJDK, redistribution is allowed for the JDK but not the JRE/JVM by themselves, so unless we either have the user's install Java on their own systems before using the auto-updater or automating that process (this can cause lots of issues, code and otherwise), we have to use the full JDK. 
As for caching Java, that would be possible, but we would have to make sure it doesn't overwrite any pre-existing java installations and we would have to fully automate the process ourselves as well.

amulet_map_editor/api/framework/update_check.py
Podshot: I forgot about that caveat, I've fixed it in my latest commits

amulet_map_editor/api/framework/update_check.py
Podshot: Fixed!

amulet_map_editor/api/framework/update_check.py
Podshot: We can do that once this is ready to be merged and I can update the URL

amulet_map_editor/api/framework/update_check.py
Podshot: I added a check for windows, what would the old logic be since this just adds the auto updater button?

amulet_map_editor/api/framework/update_check.py
```diff
@@ -18,7 +18,9 @@
 from amulet_map_editor.api.logging import log
 
 URL = "http://api.github.com/repos/Amulet-Team/Amulet-Map-Editor/releases"
-DOWNLOAD_URL = "https://github.com/Podshot/AmuletUpdater/releases/download/latest/AmuletUpdater.zip"
+
+UPDATER_DOWNLOAD_URL = "https://github.com/Podshot/AmuletUpdater/releases/download/latest/AmuletUpdater.zip"
```
gentlegiantJGC: I think it would be better if we used github's API to find the latest full release rather than hard coding the tag

Podshot: I can make that change, I used a hardcoded value since the Github API is rate limited to 60 requests an hour per computer, and the updater uses a decent portion of those requests for finding the appropriate release and the file to download, so I opted to do a hard coded URL to a release. I can make that change in a few days

gentlegiantJGC:
I have not looked too much into the legal issues (but I don't think there are any) but can we switch to using the JRE instead of the JDK? I know you said before that we can't but AdoptOpenJDK has standalone JRE releases for each platform that we can just dowload on the user's machine. They are hosted through github as well. That should reduce the file size to download and reduce the work on our end because we don't need to build/release our own Java version.
https://adoptopenjdk.net/releases.html?variant=openjdk14&jvmVariant=hotspot

Podshot:
I think if we're just downloading the .zip from their release site, then we shouldn't run into any legal issues or otherwise, the big issue was packaging the JRE with our release but if we're downloading it from their release page then I think it all checks out. I've updated the code to download and work with the JRE instead

PR #157, opened 2021-01-17
attempt to fix #133

### LINKED ISSUE #133: OSX Builds
The OSX build has issues. None of the developers are OSX users so we don't really know what we are doing and have no way to test.

If you have access to a Mac and have some idea of what you are doing please help us out.

theLMGN:


theLMGN:
still `This program needs access to the screen. Please run with a
Framework build of python, and only when you are logged in
on the main display of your Mac.`

Issue #85, opened 2020-08-12
Localization


Amirhan-Taipovjan-Greatest-I:
**Please add other languages.**

This feature can help to get more Map Editor users.

You can make ability to add languages for Amulet Map Editor.
Community can help with localization of program via pull requests.

Ex-Ark:
+1

I know there are specialized tools that integrates with Github in order to do just that.
For example [Weblate](https://weblate.org/en/hosting/) used by [some projects](https://hosted.weblate.org/) that offers to bind translation files on the VCS (git repo) to an online interface for people to correct, add, contribute to translation of all strings.

It seems like they have a free tier but this means to have a server somewhere to host the app.
Maybe there are full hosted alternative free for open source project ?

> Community can help with localization of program via pull requests.

This could also work, but you would have to manage the review manualy & have very good contribution guidelines.

Anyway I'd be happy to help for french translation if the need arise. Feel free to ping me.

gentlegiantJGC:
This is currently implemented with files found in `amulet_map_editor/lang` however only parts of the program are currently supported for localisation.

PR #159, opened 2021-01-18
Improve localization and add chinese translation

yudonglin:
Adding Chinese translation
Still, many things need to accomplish before the translation is fully functional.
Here is some of my suggestion:
- Using JSON instead of lang (which is basicly .txt)
- A config file that allows user to select their language OR
- Allow user to select their language from the setting.


yudonglin:
#85

PR #189, opened 2021-02-23
Update README.md

cpaika:
Looks like the releases now contain a mac os x build


gentlegiantJGC:
The MacOS builds have some issues so it isn't currently advertised

cpaika:
Cool, I’ll close this out then. thanks!

PR #237, opened 2021-03-28
Improve chunk tool

gentlegiantJGC:
Added a minimum and maximum draw coordinate when in 2D mode.
Moved all strings into the lang files.


amulet_map_editor/api/opengl/camera/camera.py
```diff
@@ -217,6 +227,35 @@ def orthographic_fov(self, fov: float):
         """Set the field of view of the camera when in orthographic mode."""
         self._set_fov(Projection.TOP_DOWN, fov)
 
+    def _set_clipping(self, mode: Projection, clipping: Tuple[float, float]):
+        assert (
+            type(clipping) is tuple
```
Podshot, line 232: This would be better as `isinstance(clipping, tuple)`, same for the line below which would be `all(isinstance(c, (int, float)) for c in clipping)`

Podshot (changes requested):

PR #262, opened 2021-04-15
Fix stutter

gentlegiantJGC:
There was previously a large stuttering due to merging of chunk geometry into one large array.
This was done on the main thread which locked up rendering while it was processing.
This moves it into the thread with chunk generation.


Podshot (approved):
All looks good to me!

PR #263, opened 2021-04-15
Use isinstance() for type checking instead of type()

Podshot:
Replaces all `type(obj) is <class>` type checks with `isinstance(obj, <class>)` calls

### DIFF
--- a/amulet_map_editor/api/opengl/camera/camera.py
+++ b/amulet_map_editor/api/opengl/camera/camera.py
@@ -135,11 +135,11 @@ class Camera(CanvasContainer):
         """Set the location of the camera. (x, y, z)."""
         assert (
             len(camera_location) == 3
-        ), "camera_location must be an iterable of floats with a length of three"
+        ), "camera_location must be an iterable of three floats."
         camera_location = tuple(map(float, camera_location))
         if camera_location != self._location:
             self._reset_matrix()
-            self._location = tuple(camera_location)
+            self._location = camera_location
             return True
         return False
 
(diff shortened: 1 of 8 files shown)


gentlegiantJGC:
Yeh I agree with most of this.
My only reason (mostly just in the camera) for using `type` was to make it strict since we directly store the value in the class.
Do you want to do some conversion before storing eg for the camera location to force the internal value to be `Tuple[float, float, float]` rather than whatever was given?

amulet_map_editor/api/opengl/camera/camera.py
```diff
@@ -137,9 +134,9 @@ def location(self, camera_location: CameraLocationType):
     def set_location(self, camera_location: CameraLocationType) -> bool:
         """Set the location of the camera. (x, y, z)."""
         assert (
-            type(camera_location) in (tuple, list)
+            isinstance(camera_location, (tuple, list))
             and len(camera_location) == 3
-            and all(type(v) in (int, float) for v in camera_location)
+            and all(isinstance(v, (int, float)) for v in camera_location)
         ), "format for camera_location is invalid"
```
gentlegiantJGC, line 140: for example this could just be changed to
```py
assert len(camera_location) == 3, "camera_location must be a three length iterable of floats"
camera_location = tuple(float(v) for v in camera_location)
```
and `tuple(camera_location)` can be replaced with just `camera_location`

gentlegiantJGC: Doing it this way will allow through any three length iterable who's values can be converted to floats while making the internal format strict.
This would allow you to pass a numpy array to this value and well as tuples/lists of ints/floats

Podshot: I think that's a good approach

gentlegiantJGC (approved):

PR #268, opened 2021-04-18
Improv language support

ALTB0ne:
I improved the language support a bit (43 new entries, 3 modifications).

I basically just added `lang.get()` in some places, and added their corresponding entries in `en.lang`. I kept it simple and didn't change any of the code logic, so this may not be optimal in some places depending on where the original text was. (in `__init__` files for example)

This means you may not want some of the changes made here in favor of a more in-depth code refactoring (before adding the corresponding language support).

_Black has been used for formatting in the last commit.
`test_lang.py` has been run, and changes were tested manually._

### The changes:

- Moved some colons characters inside their lang entries for better customization (those are the 3 modifications, and the only case were I modified other lang files)
- Added some missing localization here and there (Discord button, some dialogue titles, ...)
- Localized some message boxes
- Localized update checker
- Added missing calls for tab names (lang entries did already exist)
- Localized the menu bar
- Localized the loading screen of the 3D editor's canvas


amulet_map_editor/api/framework/amulet_ui.py
```diff
@@ -83,13 +83,15 @@ def __init__(self, parent):
...
-        # menu_dict.setdefault('&File', {}).setdefault('system', {}).setdefault('Create World', lambda: self.world.save())
-        menu_dict.setdefault("&File", {}).setdefault("exit", {}).setdefault(
-            "Quit", lambda evt: self.Close()
+        menu_dict.setdefault(f"&{lang.get('menu_bar.file.menu_name')}", {}).setdefault(
+            "system", {}
+        ).setdefault(
+            lang.get("menu_bar.file.open_world"), lambda evt: self._show_open_world()
         )
+        # menu_dict.setdefault(f"&{lang.get('menu_bar.file.menu_name')}", {}).setdefault('system', {}).setdefault('Create World', lambda: self.world.save())
+        menu_dict.setdefault(f"&{lang.get('menu_bar.file.menu_name')}", {}).setdefault(
```
gentlegiantJGC, line 92: The "&" symbol should be part of the localised string.
This is a special character in this context that is used as a shortcut when the alt key is pressed.
"&File" translates to "File" in the menu and when you press atl+f it will open the file menu.
Leaving this as the hard coded first character may lead to multiple menus attached to the same key codes.

More info can be found here
https://docs.wxpython.org/wx.MenuItem.html#wx.MenuItem.SetItemLabel

amulet_map_editor/api/framework/pages/world_page.py
```diff
@@ -9,13 +9,15 @@
...
+from amulet_map_editor import programs, log, lang
 from amulet_map_editor.api.datatypes import MenuData
 from amulet_map_editor.api.framework.pages import BasePageUI
 from amulet_map_editor.api.framework.programs import BaseProgram, AboutProgram
 
 _extensions: List[Tuple[str, Type[BaseProgram]]] = []
-_fixed_extensions: List[Tuple[str, Type[BaseProgram]]] = [("About", AboutProgram)]
+_fixed_extensions: List[Tuple[str, Type[BaseProgram]]] = [
+    (lang.get("program_about.tab_name"), AboutProgram)
+]
```
gentlegiantJGC, line 20: I think we need to find a better way to localise program names to avoid conflicts

amulet_map_editor/programs/edit/edit.py
```diff
@@ -140,42 +140,56 @@ def _check_close_world(self) -> bool:
...
 
-        menu.setdefault("&Edit", {}).setdefault("history", {}).update(
+        menu.setdefault(f"&{lang.get('menu_bar.edit.menu_name')}", {}).setdefault(
+            "history", {}
+        ).update(
             {
-                "Undo\tCtrl+z": lambda evt: self._canvas.undo(),
-                "Redo\tCtrl+y": lambda evt: self._canvas.redo(),
+                f"{lang.get('menu_bar.edit.undo')}\tCtrl+z": lambda evt: self._canvas.undo(),
+                f"{lang.get('menu_bar.edit.redo')}\tCtrl+y": lambda evt: self._canvas.redo(),
```
gentlegiantJGC, line 155: These should be `program_3d_edit.menu_bar._____` because they are specific to the 3d editor

gentlegiantJGC (changes requested):
This looks pretty good so far.
The only changes that need making are
- the "&" symbol in all the menubars should be part of the localised string
- menubar keys should have `program_3d_edit.` prepended

I will look into the program naming strings.

Just so you are aware we do not currently have a licence applied to this project so any code contributions you make are made with the agreement that the code can be distributed under whatever licence we eventually choose.

ALTB0ne:
Changes done:

- Move `&` inside translations;
- Prepend menubar keys with `program_3d_edit.`:
  - Except `menu_bar.options.menu_name` & `menu_bar.help.menu_name`, since they're menu _names_ which _could (?)_ be used by other subprograms (other keys in _options_ & _help_ menu tabs have been prepended, because they are specific to the 3d editor);
- Merge Chinese translations pull request (thanks yudonglin);
- Add French translations.

I didn't touch the part in `amulet_map_editor/api/framework/pages/world_page.py` ln 18-20.

I'm aware for the licence thing, no problem.

_Black has been used for formatting again. `test_lang.py` is fine too._

amulet_map_editor/api/framework/pages/world_page.py
gentlegiantJGC: Looking at this again program names do not need to be unique so this is fine.
I was also going to look into storing the name with the class but this is fine.

gentlegiantJGC (approved):

PR #334, opened 2021-06-22
Catch errors and display them to the user if a sub-program failed to load

### LINKED ISSUE #308: [Error 13] Permission denied: "resource_packs/readme.txt"
I'm getting this error when loading a world and I have been trying to find a solution to fix this from other bugs, but no luck so far. It doesn't crash the software but it will not load the world.

Here is a screenshot on what the error says
Amulet Bug

If there is anyway you could help with this bug, I would greatly appreciate it.

gentlegiantJGC:
fixes #308 
Previously if there was an error while enabling or disabling a sub-program there would be a silent error.
This shows that error to the user via a dialog so that they are aware of it.
Once they click okay the sub-program will be destroyed.


Podshot (approved):
Merging now!

PR #336, opened 2021-06-23
improve german translations

Kraballa:
Hi,
native german here. Some translations sounded like they were done through google translate so I updated them a little bit.


gentlegiantJGC (approved):
Looks good

PR #337, opened 2021-06-23
Added a warning dialog about installing certificates on macOS

gentlegiantJGC:
When running from source on macOS you need to run the install certificates file.
This is mentioned in the installer but a number of users has missed this and the error that is given is a little cryptic.
Hopefully this should make it more clear to those users.
Split up the lang entries to make it easier to add more errors


Podshot (dismissed):

gentlegiantJGC:
I am not sure if a bug report was ever opened. It was reported on the discord and I have seen the issue before.
https://discordapp.com/channels/324647192583340043/498974201416253443/856970518431531018

Podshot:
Ah ok, it's generated by the standard library when we download the resource pack. Looks good to me then!

Podshot (approved):

PR #340, opened 2021-06-23
Added a loading bar when closing worlds

### LINKED ISSUE #321: Program hangs when closing large Bedrock world
**Edition	Windows 10 Pro
Version	20H2
Compilação do SO	19042.1052
Architecture: X64
Python: 3.9+
CPU	AMD Ryzen 5 2400G with Radeon Vega Graphics       3.60 GHz
RAM installed 16,0 GB**

Whenever I open my map in 3d editor (Bedrock edition), when I save the changes and click to close the program, it stops working, as in the attached image. This always happens when I make a change to my map and try to exit the program normally right away.

**how to reproduce**

Open the amulet, load a bedrock world in version 1.17.
Open the 3d editor and then close the program
It will stop working for a long time.

image

gentlegiantJGC:
Some worlds take a while to close (particularly Bedrock which runs the compaction logic on close) and the program will hang while this is happening.
Added a dialog to reassure the user that something is happening and the program has not crashed.
This should fix #321 
Fixed a recursion issue when closing from the button in the canvas.
Fixed a crash when closing from the button in the canvas and saving.

### DIFF
--- a/amulet_map_editor/api/framework/pages/world_page.py
+++ b/amulet_map_editor/api/framework/pages/world_page.py
@@ -123,7 +123,26 @@ class WorldPageUI(wx.Notebook, BasePageUI):
         Check is_closeable before running this"""
         for ext in self._extensions:
             ext.close()
+        dialog = None
+
+        def init_dialog():
+            nonlocal dialog
+            dialog = wx.ProgressDialog(
+                "Closing World",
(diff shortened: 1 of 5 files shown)


Podshot (approved):
Looks all good to me, merging it now!

PR #348, opened 2021-06-27
Fixed UI issue when a plugin is reloaded after an error was caused by UI code

### LINKED ISSUE #346: Issues reloading plugin UI's when having syntax error.
I noticed a bug while writing some plugins.

When a plugin has a UI error, you undo your changes to the plugin that caused the issue,
you no longer able to reload the plugin with the reload button after all other plugins' UI's will no longer display correctly.

1. Create a plugin with a syntax error.
2. Load Amulet in any version.
3. Load any random world.
4. Load the operation that has the error.
5. Notice no UI in the screenshot is rendering for the working plugin "Set LootTables"
6. Try reloading the plugin after fixing the issue in the plugin, it will still not fix the UI rendering bug until you reload Amulet.

**Expected behavior is that the reload button would reload the plugins and their UIs. One plugin messing up shouldn't affect all other plugins.**

image
image

 - OS: Windows 10
 - Version 8.8

My thoughts on this are that syntax errors/issues shouldn't affect the running of the entire program and cause it to not display the UI correctly for other working plugins.

The error that I saw in the console when it was happening.

`RuntimeError: wrapped C/C++ object of type SimpleChoice has been deleted`

Podshot:
This should fix #346

### DIFF
--- a/amulet_map_editor/programs/edit/api/ui/tool/base_operation_choice.py
+++ b/amulet_map_editor/programs/edit/api/ui/tool/base_operation_choice.py
@@ -84,9 +84,14 @@ class BaseOperationChoiceToolUI(wx.BoxSizer, BaseToolUI):
             operation = self._operations[operation_path]
             if self._active_operation is not None:
                 self._active_operation.disable()
-                if isinstance(self._active_operation, wx.Window):
+                if (
+                    isinstance(self._active_operation, wx.Window)
+                    and self._active_operation
+                ):
                     self._active_operation.Destroy()
-                elif isinstance(self._active_operation, wx.Sizer):
+                elif isinstance(
+                    self._active_operation, wx.Sizer
+                ) and self._operation_sizer.GetItem(self._active_operation):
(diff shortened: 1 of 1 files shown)


gentlegiantJGC:
This only partly solves the issue and I don't think it is the correct way to do it.
I am not sure what is the proper way to solve this.

The root of the issue is that the operation UI fails to load.
As a result the new `self._active_operation` is not set so when you reload it tries to destroy the old operation that has already been destroyed.
What we need to do is try loading the UI and if that fails remove any UI that has been created.
Once that is done remove the failed operation from the list, report the error to the user and reload the previous operation.

The main issue with this is that the operation is being loaded into the canvas and is only dropped into the sizer once it is fully created but that code is not running because of the exception.

The code you modified should be changed to just destroy everything in that sizer.
I will look if there is a way to delete orphan UI elements

gentlegiantJGC:
I have just created #350 with a proper fix.

PR #349, opened 2021-06-27
Implemented a Find Block operation

Podshot:
This adds a new operation that finds a specific block (or all substates of a given block's namespaced name) and then reports the results in a tree control dialog with all of the coordinates

Example of the find block results dialog:
find_block_results

### DIFF
--- a/amulet_map_editor/programs/edit/plugins/operations/stock_plugins/operations/find_block.py
+++ b/amulet_map_editor/programs/edit/plugins/operations/stock_plugins/operations/find_block.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
+
+from functools import reduce
 from typing import TYPE_CHECKING
 
 import numpy
 import wx
+
+from operator import add
 
 from amulet_map_editor.api.wx.ui.block_select import BlockDefine
 from amulet_map_editor.api.wx.ui.simple import SimpleDialog
@@ -17,6 +21,8 @@ if TYPE_CHECKING:
 class ResultDialog(SimpleDialog):
     def __init__(self, parent, block_location_info):
         super(ResultDialog, self).__init__(parent, "Block Locations")
+
+        self._block_location_info = block_location_info
         self.Freeze()
 
         self.sizer.Add(
@@ -27,21 +33,30 @@ class ResultDialog(SimpleDialog):
             wx.ALL | wx.ALIGN_CENTRE_HORIZONTAL,
             5,
         )
-
-        tree = wx.TreeCtrl(self, style=wx.TR_DEFAULT_STYLE | wx.TR_HIDE_ROOT)
-        _root = tree.AddRoot("__root__")
-
-        for block in block_location_info:
-            block_root = tree.AppendItem(
-                _root, f"{block} - {len(block_location_info[block])} locations"
-            )
-            for coordinates in block_location_info[block]:
-                tree.AppendItem(block_root, str(coordinates))
+        self._tree = tree = wx.TreeCtrl(
+            self, style=wx.TR_DEFAULT_STYLE | wx.TR_HIDE_ROOT
+        )
+        self._root = tree.AddRoot("__root__")
 
         self.sizer.Add(tree, 1, wx.ALL | wx.EXPAND, 5)
 
         self.Layout()
         self.Thaw()
+
+    def build_results(self):
+        count = 0
+        iter_count = reduce(add, map(len, self._block_location_info.values()))
+
+        for block in self._block_location_info:
+            block_root = self._tree.AppendItem(
+                self._root,
+                f"{block} - {len(self._block_location_info[block])} locations",
+            )
+
+            for coordinates in self._block_location_info[block]:
+                self._tree.AppendItem(block_root, str(coordinates))
+                count += 1
+                yield count / iter_count, "Building results dialog..."
 
 
 class FindBlock(SimpleOperationPanel):
@@ -116,7 +131,7 @@ class FindBlock(SimpleOperationPanel):
                     )
                 )
             count += 1
-            yield count / iter_count
+            yield count / iter_count, "Searching blocks..."
 
         result = {
             world.translation_manager.get_version(
@@ -127,6 +142,7 @@ class FindBlock(SimpleOperationPanel):
         }
 
         dialog = ResultDialog(self.parent, result)
+        yield from dialog.build_results()
         dialog.Show()
 
     def _find_multiple_blockstates(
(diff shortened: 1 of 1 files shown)


amulet_map_editor/programs/edit/plugins/operations/stock_plugins/operations/find_block.py
```diff
@@ -0,0 +1,223 @@
...
+
+if TYPE_CHECKING:
+    from amulet.api.level import BaseLevel
+    from amulet.api.selection import SelectionGroup
+    from amulet.api.data_types import Dimension
+
+
+class ResultDialog(SimpleDialog):
+    def __init__(self, parent, block_location_info):
+        super(ResultDialog, self).__init__(parent, "Block Locations")
```
gentlegiantJGC, line 23: You can just do `super().__init__(parent, "Block Locations")`

amulet_map_editor/programs/edit/plugins/operations/stock_plugins/operations/find_block.py
```diff
@@ -0,0 +1,223 @@
...
+
+            for coordinates in self._block_location_info[block]:
+                self._tree.AppendItem(block_root, str(coordinates))
+                count += 1
+                yield count / iter_count, "Building results dialog..."
+
+
+class FindBlock(SimpleOperationPanel):
+    def __init__(self, parent, canvas, world, options_path):
+        SimpleOperationPanel.__init__(self, parent, canvas, world, options_path)
```
gentlegiantJGC, line 64: You can just do `super().__init__(parent, canvas, world, options_path)`

amulet_map_editor/programs/edit/plugins/operations/stock_plugins/operations/find_block.py
```diff
@@ -0,0 +1,223 @@
...
+                    self._block_define.properties,
+                ),
+                "find_substates": self._get_find_all_substates(),
+            }
+        )
+
+    def _find_single_blockstate(self, world, dimension, selection):
+        target = world.block_palette.get_add_block(self._get_target_block())
+
+        iter_count = len(list(world.get_chunk_slice_box(dimension, selection, False)))
```
gentlegiantJGC, line 117: I know I have done this a bit but there is a quicker way. I should look into fixing the other places.
`len(list(world.get_coord_box(dimension, selection, False)))`
This should get the same length (or at least close) without having to pre-load all the chunks.

amulet_map_editor/programs/edit/plugins/operations/stock_plugins/operations/find_block.py
```diff
@@ -0,0 +1,223 @@
...
+
+        self._block_define = BlockDefine(
+            self,
+            world.translation_manager,
+            wx.VERTICAL,
+            *(
+                options.get("target_block_options", [])
+                or [world.level_wrapper.platform]
+            ),
+            show_pick_block=True,
```
gentlegiantJGC, line 80: You haven't hooked up the pick block tool to the render.
Check out the fill operation for an example of how to do that.

gentlegiantJGC:
I like what you have done. I have commented on some minor code issues.
A few more issues.
The output x and z coordinates are wrong. It looks like they need the chunk base location added to them.
It would be really nice if you could click on one of the coordinates in the UI and get teleported to the location in the world. I don't think that should be too difficult to do.

Sorry about the next bit. It is a bit rambly but I don't know how to explain it concisely
One issue that would need quite a bit of rewriting is that you are doing the matching in the universal format. I don't have an issue with this but there are some cases where it might give some obscure results.
The better option would be to convert each blockstate from the universal format and see if it matches the block defined in the version format.

Example:
Block to find Bedrock 1.17 `minecraft:oak_stairs[upside_down_bit=0b,weirdo_direction=0]`
When converted to universal `minecraft:oak_stairs[material=oak,half=bottom,facing=east,shape=straight]`

Universal blocks in the world
`minecraft:oak_stairs[material=oak,half=bottom,facing=east,shape=straight]`
`minecraft:oak_stairs[material=oak,half=bottom,facing=east,shape=inner_left]`
`minecraft:oak_stairs[material=oak,half=bottom,facing=east,shape=inner_right]`

If you match in universal only the first will match. The blocks in world converted to Bedrock 1.17 will all be the same as the first blockstate.

Podshot:
For finding locations of a single blockstate I compare using the universal format, but for finding all the substates of a block I first convert everything to the versioned block, are you saying that I should do the same for searching when doing a single blockstate?

gentlegiantJGC:
In the multiple state you are converting to the world version. Y
 getting the universal block from the UI and then converting it to the world version. You are then converting the blocks in the world to that

gentlegiantJGC:
You should should be converting to the version that is chosen in the BlockDefine UI not the world version that way you are matching in whatever format the user defined the block in.

Sorry about closing the issue. I think I hit shift + enter on autopilot to do a new line like in discord which sent the message and closed the issue

amulet_map_editor/programs/edit/plugins/operations/stock_plugins/operations/find_block.py
```diff
@@ -0,0 +1,223 @@
...
+        # is missing blockstates when the results window is shown
+        blocks_to_find = blocks_to_find.union(
+            filter(
+                lambda b: b[1].namespaced_name == versioned_block_name,
+                map(
+                    lambda b: (b[0], block_translator.from_universal(b[1])[0]),
+                    world.block_palette.items(),
+                ),
+            ),
+        )
```
gentlegiantJGC, line 177: This up here is fine but as each chunk is loaded if new blockstates are found in that chunk they will be added to the block palette.
In the loop below you should check if the palette size has changed and scan the new blocks

amulet_map_editor/programs/edit/plugins/operations/stock_plugins/operations/find_block.py
```diff
@@ -0,0 +1,223 @@
...
+
+        iter_count = len(chunk_slices)
+        iter_count *= len(blocks_to_find)
+        count = 0
+
+        results = {block.full_blockstate: [] for _, block in blocks_to_find}
+
+        for internal_id, block in blocks_to_find:
+            extend_func = results[block.full_blockstate].extend
+            for chunk, slices, _ in chunk_slices:
```
gentlegiantJGC, line 189: These loops would be better the other way around so that you are only loading each chunk once and doing all the work on that chunk in one go.

gentlegiantJGC:
It is also worth mentioning that I am working on a UI that will allow the user to pick one or more values for each property.
It would probably make sense for this to use that UI which will allow a lot more flexibility for the users.
Eg if the user wanted to find all upside down stair blocks regardless of their rotation

PR #350, opened 2021-06-27
Handled Operation UI errors and remove the UI

### LINKED ISSUE #346: Issues reloading plugin UI's when having syntax error.

gentlegiantJGC:
If an operation fails to load for some reason this will destroy the created UI and restore everything back to the starting state.
Once this is done the traceback is show to the user.
The user can modify the code to fix the issue and reload the operations and it should appear.

This should fix #346

### DIFF
--- a/amulet_map_editor/programs/edit/api/ui/tool/base_operation_choice.py
+++ b/amulet_map_editor/programs/edit/api/ui/tool/base_operation_choice.py
@@ -1,13 +1,13 @@
 import wx
 from typing import TYPE_CHECKING, Optional
+import traceback
 
 from amulet_map_editor import log
 from amulet_map_editor.api.image import REFRESH_ICON
 from amulet_map_editor.api.wx.ui.simple import SimpleChoiceAny
+from amulet_map_editor.api.wx.ui.traceback_dialog import TracebackDialog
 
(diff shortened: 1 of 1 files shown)


Podshot (approved):
Looks all good to me! Merging it now!

PR #358, opened 2021-06-30
Fix requirements

### LINKED ISSUE #335: Better Linux setup
Currently, Linux user need to manually install wxPython in order to use Amulet.
This is because the default wxPython won't work for every linux os. 
So, some linux os can use pip to install it, and some other need to build it themselves

As mentioned in #281 , it could be possible to generate a specific requirement.txt for each OS in setup.py   
(It's possible to get the name of the Linux OS using `os.uname()[1]`)

### LINKED ISSUE #354: Freeze requirements in the build system
The build system should freeze the requirements and replace the requirements.txt
This will make sure that if a user updates the program it will always get the required dependency rather than leaving the old version.
It also means that if we accidentally break something in a future release of the dependencies the old versions will still work.

gentlegiantJGC:
fixes #335
fixes #354

Makes the wxpython requirement required on linux. This should make it install from source. There was probably a reason why this was done like this but I can't remember what it was. Added wheel on linux because I think installing wx from source requires it.

Cleaned up the manifest.in file

Froze the first party libraries before uploading to pypi to ensure that users have the latest versions.


gentlegiantJGC:
This has an issue that the source is modified so the version is getting an extension which is not valid on pypi.

Not sure how to get around that.

gentlegiantJGC:
I think that should fix it

Podshot (approved):
Looks all good to me, merging this now!

PR #362, opened 2021-07-03
Updated location of wx.VERSION for linux

VolchaH:
When starting on linux, amulet almost immediately crashes, returning this as part of the traceback:

        if sys.platform == "linux" and wx.__version__.VERSION >= (4, 1, 1):
    AttributeError: 'str' object has no attribute 'VERSION'

It seems that wx moved the tuple version information, which I updated to match. The proposed fix has been tested with Python 3.7 on Debian 10.1, and with Python 3.9 on a recent Debian Testing.

### DIFF
--- a/amulet_map_editor/__main__.py
+++ b/amulet_map_editor/__main__.py
@@ -13,7 +13,7 @@ from amulet_map_editor.api.framework import AmuletApp
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


gentlegiantJGC:
Nice catch. It looks like the variable in the main init is shadowing the module. `VERSION` is accessible there because it is indirectly imported twice which I missed.

gentlegiantJGC (approved):

PR #369, opened 2021-07-06
Moved imports in __main__ into a try except block

### LINKED ISSUE #368: Improve instant crash
If part of Amulet does not exist it will display the error and then crash immediately.
The module importing should also be in a try except block to handle these errors.

gentlegiantJGC:
If part of Amulet was missing, the error would happen outside the existing try except block.
This moves the imports into a try except block so that if an error does occur it can be reported correctly.
This fixes #368


Podshot (approved):
Looks all good to me!

PR #378, opened 2021-07-15
Fixed update check to now support beta/alpha/nightly releases

Podshot:
Implemented a better update checking system, that now correctly identifies our versioning scheme.

Changes:
- Checking whether an update is valid is now based off of a version hierarchy (IE: Full releases only notify will notify users of new full releases, running a beta build will notify users of new beta or full releases, etc.)
- New nightly builds are correctly identified
- More efficient and accurate newest release finding
  - The old finding code sometimes would just find the next newest version, not the actual newest version
- Rudimentary checking for when running from source with a git installation to not show the update dialog


gentlegiantJGC:
This looks good. I have made some changes and created #382 for you to look over them and pull them

gentlegiantJGC (approved):
Looks good

PR #382, opened 2021-07-27
Improv fix update checking

gentlegiantJGC:
I have added a better regex match that matches every version we have released and every format we will release.
I don't think it will break anything you have changed but double check it.
I also converted the release stages to an enum

### DIFF
--- a/amulet_map_editor/api/framework/update_check.py
+++ b/amulet_map_editor/api/framework/update_check.py
@@ -13,7 +13,24 @@ from amulet_map_editor import lang
 
 URL = "http://api.github.com/repos/Amulet-Team/Amulet-Map-Editor/releases"
 VERSION_REGEX = re.compile(
-    r"^v?(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)((b(?P<beta>\d+)(\.dev(?P<devnum>\d{12}))?)|(a(?P<alpha>\d+)))?(.*\.g(?P<commit_hash>[a-z\d]{7}))?"
+    r"^"  # Match the beginning
+    r"v?" # Match an optional "v" at the start.
+    r"(?P<major>\d+)"  # Major version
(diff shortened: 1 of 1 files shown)


Podshot:
I don't think making the patch number optional is a good idea, since that is decently important part of our semantic versioning (while the commit hash/dirty attributes are extra/running from source only). I like the explanation of the regular expression, but what do you think about making the documentation on that a block comment before or after the `re.compile()` call? I feel the breaking up of the regular expression line-by-line is a bit awkward. 

I didn't focus on being able to match against all of our release because this change won't be retroactive really, it just needs to work with/match the first release version it will be included in and all the releases after that.

gentlegiantJGC:
Some of the modifications were to make it more compliant with PEP 440 which the version system is based on but we only use a sub-set of the format.
If the patch version is not included it is equivalent to 0 eg `8.1 == 8.1.0`. I changed that in the code so that if it is not included it sets it to 0. This is how PEP 440 outlines the versioning.

`[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`

I though the regex was rather complex on one line which is why I broke it up and documented it. I wanted to add some indentation to better show the grouping but I couldn't work out how to do that.

Podshot:
I know that we currently have versioneer using the PEP440 versioning scheme, but I think we should have a stricter subset for our version scheme. I don't think we really need to have any number of `.*` after the major/minor/patch numbers, since anything after patch should just be another patch, not a new subversion of the previous patch release. 

As the code currently is implemented, anything after the patch number is not even picked up, so `1.1.1` would be equated with `1.1.1.1`, which if we were to support any number of additional subversions would just add unneeded complexity in my opinion. 

I've reworked the regular expression documentation as well as the regular expression to be back on one line and include the tweaks you made to it but removed the part that would match `(.N)*`

gentlegiantJGC:
Looks good. Sorry I forgot about this

Podshot (approved):

PR #386, opened 2021-08-02
Optimized images.

RDKRACZ:
Lossless compression.


gentlegiantJGC:
It looks like you have just converted the RGBA image to LA (black with alpha).
I would need to look into what issues this may cause. I can't remember how we unpack the images and if the classes that use the images expect an RGB(A) image. If not this could save a little data but all the images combined sum to less than 1MB so it isn't really a major concern.
The images can be compressed even further by first converting to LA and then to a paletted image.

PR #388, opened 2021-08-07
Improve setup logic

gentlegiantJGC:
Move to the new build system as described in PEP 517 and 518
https://www.python.org/dev/peps/pep-0517/
https://www.python.org/dev/peps/pep-0518/

This is the more modern way of building projects.
We are going to need to add cython logic to the build process so it made sense to clean this up first.

cleaned up setup.py

- Reworked the build workflow:
  - Switched to a list based matrix for more flexibility.
  - Added an x86 build for windows.
  - Removed the macOS build until we work out how to handle it properly.
  - Set bash as the default shell
  - Streamlined it quite a lot


Podshot (approved):
Looks all good to me! Merging this now!

PR #401, opened 2021-08-20
Improve NBT editor

Podshot:
Reworked the NBT editor to use an abstracted representation of a NBT structure instead of actual NBT objects. This allows for much cleaner handling of adding/removing items from container types (`TAG_Compound`/`TAG_List`), as well as simplifying the UI code to reduce the need to handle edge cases.

Changes:
- Input NBT objects are parsed to a representation format that keeps track of the parent containing it
- Tree building logic is less complex and cleaner
- NBT editing/adding dialogs are separate from each other
- Removed the need for a callback (current NBT state can be accessed via the `.nbt` property)

### DIFF
--- a/amulet_map_editor/api/wx/ui/nbt_editor.py
+++ b/amulet_map_editor/api/wx/ui/nbt_editor.py
@@ -161,7 +161,7 @@ class NBTEditor(simple.SimplePanel):
         tag_name, tag_obj = self.tree.GetItemData(evt.GetItem())
 
         menu = self._generate_menu(
-            isinstance(tag_obj, (MutableMapping, MutableSequence))
+            isinstance(tag_obj, (MutableMapping, MutableSequence, nbt.NBTFile))
         )
         self.PopupMenu(menu, evt.GetPoint())
         menu.Destroy()
@@ -265,6 +265,15 @@ class EditTagDialog(wx.Frame):
     GRID_ROWS = 3
     GRID_COLUMNS = 4
 
+    DEFAULT_VALUES = {
+        "TAG_Int": "0",
+        "TAG_Long": "0",
+        "TAG_Short": "0",
+        "TAG_Byte": "0",
+        "TAG_Float": "0.0",
+        "TAG_Double": "0.0"
+    }
+
     def __init__(
         self, parent, tag_name, tag, tag_types, create=False, save_callback=None
     ):
@@ -345,11 +354,18 @@ class EditTagDialog(wx.Frame):
 
     def value_changed(self, evt):
         tag_value = evt.GetString()
-        self.value_field.ChangeValue(str(self.data_type_func(tag_value)))
+        try:
+            self.value_field.ChangeValue(str(self.data_type_func(tag_value)))
+        except ValueError:
+            self.value_field.ChangeValue(self.DEFAULT_VALUES[tag_value])
 
     def handle_radio_button(self, tag_type, evt):
         for rd_btn in self.radio_buttons:
             rd_btn.SetValue(rd_btn.nbt_tag_class == tag_type)
+        if tag_type in ("TAG_Compound", "TAG_List"):
+            self.value_field.Disable()
+        elif not self.value_field.Enabled:
+            self.value_field.Enable()
         self.change_tag_type_func(tag_type)
 
     def change_tag_type_func(self, tag_type):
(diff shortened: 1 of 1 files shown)


amulet_map_editor/api/wx/ui/nbt_editor.py
```diff
@@ -1,26 +1,197 @@
...
+        raise NotImplementedError
+
+    def build_nbt(self):
+        raise NotImplementedError
+
+
+class ParsedNBTValue(ParsedNBT):
+    __slots__ = ParsedNBT.__slots__ + [
+        "_value",
+    ]
```
gentlegiantJGC, line 72: You can just re-define `__slots__` as ("_value",) and it will extend the parent

amulet_map_editor/api/wx/ui/nbt_editor.py
```diff
@@ -1,26 +1,197 @@
...
+
+        return container
+
+
+def parse_nbt(
+    _nbt: Union[nbt.NBTFile, nbt.TAG_Compound, MutableMapping],
+    parent: ParsedNBTContainer = None,
+):
+    for key, value in _nbt.items():
+        if isinstance(value, MutableMapping):
```
gentlegiantJGC, line 167: I think this might break with the python version of the nbt library. Please can you test all of this with the python version as well

gentlegiantJGC:
There are still some issues that need ironing out but it does look good.
I re-added the old example nbt because you had it linked to a file on your PC but the old NBT cannot be displayed.
For some reason it cannot display non-container data that is either the root or in the first level.
This NBT viewer should be able to display any valid nbt or NBTFile

gentlegiantJGC:
The tree should really have one root node and all of the NBT should be a sub-entry under that node.

gentlegiantJGC:
Bugs:
These things cannot be displayed
```py
nbt.TAG_Compound({})  # This shows nothing
nbt.TAG_String("hello")  # This causes a crash
nbt.TAG_Compound({"str": nbt.TAG_String("hello")})  # This causes a crash
```

I cannot edit arrays.
Not really a bug but I intuitively double click on items to edit them. I don't think that should be too difficult to add.
wx assertion error when deleting the "root" compound tag (note this is not actually the root of the NBT)
The add nbt tag window is not big enough by default to see the save/cancel buttons
The icons in the new tag window could be closer to the radio button they are associated with. It currently looks like they are associated to the one to the right not the left.
Creating a tag in a list asks me for a name. Tags in lists do not have names
I am able to create tags of different types in a list. NBT can only have one type in a list
I can create two tags in a compound with the same name

amulet_map_editor/api/wx/ui/nbt_editor.py
```diff
@@ -585,7 +585,7 @@ def _save(self, _):
 
 if __name__ == "__main__":
     level_dat = nbt.load(
-        r"C:\Users\gotharbg\Documents\Python Projects\Amulet-Core\tests\worlds_src\java_vanilla_1_13\level.dat"
+        b"\x0A\x00\x0B\x68\x65\x6C\x6C\x6F\x20\x77\x6F\x72\x6C\x64\x08\x00\x04\x6E\x61\x6D\x65\x00\x09\x42\x61\x6E\x61\x6E\x72\x61\x6D\x61\x00"
```
Podshot, line 588: This is an invalid NBT bytestring, since the `TAG_String` is not wrapped in a `TAG_Compound` which is the reason this string crashes the UI

gentlegiantJGC: And a tag string can be standalone in the binary format it just isn't used

Podshot: I thought all valid NBT had to be wrapped in a `TAG_Compound`? The NBTFile object shouldn't be acting as that `TAG_Compound`

Podshot: I'm referencing this page for whether the byte string is valid NBT or not: https://wiki.vg/NBT#Specification

gentlegiantJGC: I think you are misreading it a bit. I assume you are referring to this line:
`Every NBT file will always implicitly be inside a tag compound, and also begin with a TAG_Compound`. The main word being `implicitly`

Normally a dictionary will own the keys and store the values under those keys. Binary NBT is defined a bit backwards. 
Instead of the compound tag owning the keys, the tag itself owns the keys. This is because the root tag also has a name.

The implicit compound tag is there to store the leading name however it is only used in Java level.dat files. Everywhere else it is blank. The NBTFile class is just there to handle the leading name if it isn't blank.

It is a bit complex. I am not sure if I explained it that well

gentlegiantJGC: This is the code you will need to read a bedrock level.dat file
```py
with open(path, "rb") as f:
        f.read(8)
        dat = f.read()
        level_dat = nbt.load(
            dat,
            little_endian=True
        )
```

gentlegiantJGC: Also while doing some debugging I realised our nbt library cannot de-serialise binary root tags that are not compound tags. This isn't really an issue though since I don't think it is ever used but they can be constructed either from SNBT or by directly using the classes so the viewer should be able to display them.

PR #412, opened 2021-09-09
Updated licence info

gentlegiantJGC:
Added LICENCE.md file
Added MIT licence to the stock operations and the language files


gentlegiantJGC:
Once sorted out we should look into setting up this or something like it
https://cla-assistant.io/

rster2002:
With how the LICENSE is currently set up, the amulet editor wouldn't be considered open source anymore as by the open source initiative's definition: https://opensource.org/osd just be aware of that.

PR #424, opened 2021-09-23
Fixed update check

gentlegiantJGC:
The old code was recommending an older full release over the newer beta.


Podshot (approved):
Looks all good to me! Merging this now

PR #434, opened 2021-09-28
Fix build version

### LINKED ISSUE #429: New build system does not bake the requirements
I am seeing the following error whenever Amulet attempts to load one of my worlds.

```
2021-09-24 01:49:01,917 - amulet_core - ERROR - Error loading chunk -11 65 minecraft:overworld
  File "/home/user/.virtualenvs/mcbeedit/lib/python3.7/site-packages/amulet/api/wrapper/format_wrapper.py", line 433, in _safe_load
    return meth(*args)
  File "/home/user/.virtualenvs/mcbeedit/lib/python3.7/site-packages/amulet/api/wrapper/format_wrapper.py", line 481, in _load_chunk
    interface, dimension, cx, cz, raw_chunk_data
  File "/home/user/.virtualenvs/mcbeedit/lib/python3.7/site-packages/amulet/level/formats/leveldb_world/format.py", line 242, in _decode
    return interface.decode(cx, cz, raw_chunk_data, (bounds[0][1], bounds[1][1]))
  File "/home/user/.virtualenvs/mcbeedit/lib/python3.7/site-packages/amulet/level/interfaces/chunk/leveldb/base_leveldb_interface.py", line 111, in decode
    chunk.blocks, chunk_palette = self._load_subchunks(subchunks)
  File "/home/user/.virtualenvs/mcbeedit/lib/python3.7/site-packages/amulet/level/interfaces/chunk/leveldb/base_leveldb_interface.py", line 397, in _load_subchunks
    raise Exception(f"sub-chunk version {data[0]} is not known.")
Exception: sub-chunk version 9 is not known.
```

The world save is from Bedrock Dedicated Server version 1.17.31.01
Amulet version: v0.8.13

The error is reported any time Amulet attempts to process a chunk that was altered since the world was upgraded from 1.17.11. In the Amulet UI, the affected chunks appear to be void chunks with a red overlay over them.

gentlegiantJGC:
This should fix #429 

This adds an optional flag to the setup logic which enables installing all of the requirements and then freezing the first party libraries.
This means that the newest compatible version will be gotten at build time so that source users will have the same version as the compiled users. It will also mean that source users have the newest version of each library if reinstalling.
This will only be done when building the wheel file if the extra input is defined.

I tried doing this without installing the libraries but making sure they were all compatible was a bit of a nightmare so I reverted back to using pip.

Note that the build flag is defined twice due to an issue with build calling setuptools. I have opened a pull request with setuptools to try and fix the issue.


gentlegiantJGC:
I am going to take a break from this. I have it working locally but for some reason I can't work out how to set an environment variable in the workflow. If you have any ideas @Podshot let me know.

Podshot:
I'll do some investigating in a day or two, I'm wondering if it's only for the shell/commands used in the workflow or if those variables get set to the entire host which then would be accessible to python

gentlegiantJGC:
I think this is ready for you to look over now @Podshot 
I have set a build going to make sure that it works as intended.
https://github.com/Amulet-Team/Amulet-Map-Editor/releases/tag/0.8.16a0

gentlegiantJGC:
Actually just realised I broke something

gentlegiantJGC:
Okay I think that should fix it

Podshot (approved):
Looks good to me! Merging now!

gentlegiantJGC:
thanks

PR #458, opened 2021-11-11
Removed stack inspection from run_operation and made thread safe

gentlegiantJGC:
This will allow the first thread that calls it to run and hold back the second thread until the first is finished.
This uses the RLock class so that if the same thread calls the method again it will be allowed through but then throw an exception rather than get stuck acquiring the lock.


Podshot (approved):
Looks all good to me!

PR #492, opened 2021-12-09
Fix window size restoring issues

### LINKED ISSUE #467: Initial app size is too small for us beginners
When I try to convert a Bedrock world to a Java world, it looks like this in Amulet:

image

Where did the convert button go? I was lost, until I finally realized I could resize the application and find it:

image

I suggest that the initial size of the application be large enough that this confusing thing doesn't happen by default.

gentlegiantJGC:
There were a number of issues with the code that saved and restored the window size after restarting the program.

Added an `is_full_screen` attribute to the config file to track if the window is full screen or not.
Restore the window to either windowed in the same location or full screen depending on the previous state.
Fixed an issue where the window started in full screen but showed the restore down button.
Made the window start in full screen the first time the program is loaded.
The config file now stores the non-maximised size. This allows the window to be start in full screen but be restored down to the size before full screening.
Increased the default app size a bit to fix #467

### DIFF
--- a/amulet_map_editor/api/framework/amulet_ui.py
+++ b/amulet_map_editor/api/framework/amulet_ui.py
@@ -41,7 +41,7 @@ class AmuletUI(wx.Frame):
             id=wx.ID_ANY,
             title=f"Amulet V{__version__}",
             pos=wx.DefaultPosition,
-            size=wx.Size(560, 400),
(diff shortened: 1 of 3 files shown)


Podshot (approved):
Looks good to me! Merging this now!

PR #493, opened 2021-12-09
Fixed build and bumped core version

gentlegiantJGC:
Sometimes the build system caches the dependencies leading to two successive runs to sometimes use the old version.
This forces using the new version.
Bumped the core version to use the alpha versions of our core library


Podshot (approved):
Looks all good to me! Merging now!

PR #530, opened 2022-01-12
Cleaned up the pyinstaller spec file

gentlegiantJGC:
Removed the manual inclusion of the pyinstaller data.
This is now handled in the hook file added here
https://github.com/gentlegiantJGC/PyMCTranslate/commit/39805bd3721a46c5ff4fd966e2b6e21361888333
Also added in the nbt hook which was added here
https://github.com/Amulet-Team/Amulet-NBT/commit/f1a782d6963fbb7aa8c278c234d0c7454e0aacd3

This can wait to get merged until the new NBT library is ready. I think it should just ignore the hook path if it can't find it.


Podshot (approved):
Looks all good to me!

PR #570, opened 2022-02-19
Improved world closing UI

gentlegiantJGC:
Previously when the world was closed there was always a dialog that popped up to notify the user that it might take a while. This always shown even if only instantaneously.

This moves the closing logic to a new thread so that it does not block the UI and the closing notification is only shown if it takes more than 0.1 seconds to close.

This also fixes an issue where the UI would become unresponsive if the world save code ran too quickly. This was due to the close notification being present but hidden. I think this was related to dialog's parent being destroyed. Removing the parent seems to fix this.

### DIFF
--- a/amulet_map_editor/api/framework/pages/world_page.py
+++ b/amulet_map_editor/api/framework/pages/world_page.py
@@ -1,5 +1,4 @@
 import threading
-import time
 
 import wx
 from typing import List, Callable, Tuple, Type, Union, Optional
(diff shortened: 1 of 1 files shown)


Podshot (changes requested):
Instead of using a nonlocal closed boolean and `time.sleep()`, we can use `thread.join(0.1)` and a `thread.is_alive()` in the if statements after that to better check if the thread has finished

gentlegiantJGC:
That is a much better way of doing it. I have made the changes.

Podshot (approved):

PR #579, opened 2022-02-28
feat: Define entry-point

eaglesemanation:
This will automatically create executable script that runs 'main' function and add it to a folder that should be a part of PATH env var. This is useful for tools like pipx


gentlegiantJGC:
Sorry it has taken me a while to get back to you.
I have been meaning to set up a Contributor Licence Agreement form to get permission to use contributions.
Please can you confirm that.

I would also prefer that the main function was in its own module rather than being defined in the `__init__.py` file.

gentlegiantJGC:
Sorry for hijacking your pull request. I was going to change it so much that it was going to be unrecognisable so I started from the origin.
See #673

PR #592, opened 2022-03-09
Delete blocks does not create chunks

### LINKED ISSUE #550: [Bug Report] Massive Holes
## Bug Report
**Amulet deletes whole chunks instead of selected blocks.**
I've tried to cut an hole from `0 0 0 xyz`  all the way to `10k 391 -1k xyz`.

### Current Behaviour:
It did cut the blocks, but cut millions more than it needed. 

### Expected behavior:
It should have deleted block above `~ 0y ~` but instead looks like it deleted entire chunks. From sky limit down to bedrock.

### Steps To Reproduce:
Generate classic world.
Load world data in Amulet.
Select anything above 10k * 391 * 1k
Click Delete
Click Save

### Environment:
- OS: Windows 
- Minecraft Platform: Java 
- Minecraft Version: 1.18.1
- Amulet Version: 0.8.17

## Attachments
[worldgen_settings_export.json](https://drive.google.com/file/d/1CUnKTzkQ1GHJP-MyT9W6DUv4MZvA9vAs/view?usp=sharing)

### Worlds
[World](https://drive.google.com/file/d/1sLMm5ZNlhaQ2U6Vtpborp5LQJ_Dp8jvb/view?usp=sharing)

gentlegiantJGC:
Fixes #550 
Modified delete operation to only delete blocks in chunks that exists.
This means that chunks that do not exist will not be created.

### DIFF
--- a/amulet_map_editor/programs/edit/plugins/operations/stock_plugins/internal_operations/delete.py
+++ b/amulet_map_editor/programs/edit/plugins/operations/stock_plugins/internal_operations/delete.py
@@ -1,6 +1,7 @@
+from __future__ import annotations
+
 from typing import TYPE_CHECKING
 
-from amulet.operations.fill import fill
 from amulet.api.selection import SelectionGroup
 from amulet.api.block import UniversalAirBlock
 from amulet.api.data_types import Dimension, OperationReturnType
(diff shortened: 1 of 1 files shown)


Podshot (approved):
Looks good to me, merging now!

Issue #393, opened 2021-08-11
"Create Empty Chunks" button


sparks0:
**Is your feature request related to a problem? Please describe.**
Sometimes I need to create empty void areas for worlds that need two types of generation for different zones such as a world that also has a "space" section you can visit. Currently this has to be done by creating the empty chunks in MCEdit, then converting the world to Bedrock and then merging the world with any builds.

**Describe the solution you'd like**
In the 3D editor of Amulet, when "Chunk" mode is selected, a button called "Create Empty Chunks" or "Fill with Empty Chunks" should appear in the context menu under "Delete Unselected Chunks". Clicking this button would generate empty chunks with nothing but air within the selection, creating new chunks where no chunks exist, and replacing existing chunks' content with air.

**Describe alternatives you've considered**
Currently this feature technically exists as selecting an empty space with the chunk tool and pressing the "delete" key will do exactly this, but since this is a hidden feature and also I was told this is a bug and the delete key should only empty existing chunks, not create new ones, it would be good to have a button that makes the feature official and is more likely to be discovered and used by users.

*Other context*
I understand that in the 3D view in "Select" mode, the delete key makes sense to delete the blocks within the selection, however in 2D chunk mode, where it is the CHUNK that is selected not the blocks inside it, it makes more sense for the delete key to completely erase the chunk.

gentlegiantJGC:
This was implemented in the last release

PR #593, opened 2022-03-09
Added a button to create empty chunks

gentlegiantJGC:
This was suggested in #393
This button will create all chunks in the selection that do not already exist


Podshot (approved):
Looks all good to me!

PR #635, opened 2022-04-09
Added validators to the numerical inputs

gentlegiantJGC:
It was possible before to type non-numerical characters which made the input default to the minimum

Fixes #635


Podshot (approved):
Looks all good to me!

PR #637, opened 2022-04-12
Moved operation running to a thread

gentlegiantJGC:
This moves the actual operation to a new thread so that the UI does not lock up when an operation takes a long time to run.
This applies to function operations and generator operations but only the latter will display estimated time like before.
Also added the ability to cancel operations.
Modified the save logic to use the new operation running system.
Depreciated the old show_loading_dialog function. I don't know if third party code still uses this.

### DIFF
--- a/amulet_map_editor/programs/edit/api/canvas/edit_canvas.py
+++ b/amulet_map_editor/programs/edit/api/canvas/edit_canvas.py
@@ -1,7 +1,9 @@
+import warnings
+
 import wx
 from typing import Callable, TYPE_CHECKING, Any, Generator, Optional
 from types import GeneratorType
-from threading import RLock
+from threading import RLock, Thread
 
 from .base_edit_canvas import BaseEditCanvas
 from ...edit import EDIT_CONFIG_ID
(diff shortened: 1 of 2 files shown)


amulet_map_editor/programs/edit/api/canvas/edit_canvas.py
```diff
@@ -159,73 +213,103 @@ def _deselect(self):
...
                 self.world.restore_last_undo_point()
 
+                if isinstance(op.error, BaseLoudException):
+                    msg = str(op.error)
+                    if isinstance(op.error, OperationError):
+                        msg = f"Error running operation: {msg}"
+                    log.info(msg)
+                    wx.MessageDialog(self, msg, style=wx.OK).ShowModal()
+                elif isinstance(op.error, BaseSilentException):
+                    pass
```
Podshot, line 294: Even if the exception is silent and not shown to the user, I would like it to still be logged to the console window and/or log file just to make debugging easier if an issue caused by a silent exception ever arises with an operation. If it's caught and logged somewhere else that I'm not seeing then that's fine with me

gentlegiantJGC: The silent exception is there for cases where the operation does not want to create an undo point and does not want to notify the user.
An example is the copy selection operation. It isn't really an exception as much as a notification.

PR #638, opened 2022-04-12
Added numpy.import_array()

gentlegiantJGC:
This is apparently required in some situations and is better to have it than not.
I hit the error of not having it once but I cannot reproduce it.


Podshot (approved):
Looks good to me! Merge this whenever you need to

PR #647, opened 2022-04-20
Improved logging

gentlegiantJGC:
Switched to getLogger to get the logger instead of importing the object.
Added file handlers for the libraries.
Merged the file output into one file.
Made the file include the process id so that each process has its own log file.

This is dependent upon the following library pull requests.
https://github.com/Amulet-Team/Amulet-Core/pull/183
https://github.com/gentlegiantJGC/Minecraft-Model-Reader/pull/12
https://github.com/gentlegiantJGC/PyMCTranslate/pull/16

Please review and I will merge when ready.


amulet_map_editor/programs/edit/edit.py
```diff
@@ -1,5 +1,9 @@
 from typing import TYPE_CHECKING, Optional, Generator
 import webbrowser
+import logging
+
+EDIT_CONFIG_ID = "amulet_edit"
```
Podshot, line 5: This seems to already be defined earlier in the file, are one of these definitions supposed to be removed or is this intentional?

Podshot (changes requested):

amulet_map_editor/programs/edit/edit.py
gentlegiantJGC: looks like an error when I was merging the branches

gentlegiantJGC: Just fixed this. Thanks for pointing it out

Podshot (approved):
Looks all good to me in it's current state! If any changes need to be made before it's merged I can do a quick review of them

Issue #407, opened 2021-08-31
Delete key inside type boxes deletes world selection instead of typed text


qmagnet:
**Describe the bug**
When you press the delete key inside the replace operation typing boxes, it deletes whatever is inside your selection box

**To Reproduce**
1. Open world
2. Click Operations
3. Click Replace
4. Put mouse cursor inside namespace field
5. Press the delete key

**Expected behavior**
This should delete the letters inside the typing box

**Screenshots**
![image](https://user-images.githubusercontent.com/13902631/131520900-dae97e36-894f-458b-bae2-7b9bd57f7987.png)

**Desktop (please complete the following information):**
 - Windows
 - Amulet v0.8.12

gentlegiantJGC:
Currently it seems that the best action is just to block these key presses.

The issue is due to a conflict with the menu system and the widget.

I have been discussing it in this forum post
https://discuss.wxpython.org/t/menubar-spinctrl-and-copy/35968/10

and opened this bug report on the wxPython repo
https://github.com/wxWidgets/Phoenix/issues/2148

DietmarSchwertberger:
What type is the control that has the focus?
I'm working on a fix, but that's currently only for spin controls.
( https://github.com/wxWidgets/wxWidgets/pull/22397 )

gentlegiantJGC:
> What type is the control that has the focus? I'm working on a fix, but that's currently only for spin controls. ( [wxWidgets/wxWidgets#22397](https://github.com/wxWidgets/wxWidgets/pull/22397) )

I can't reproduce the issue in ComboBox or SearchCtrl in WxPython 4.1.1.
I can only reproduce it in SpinCtrl

@qmagnet are you still getting this issue in the namespace and block name fields?

DietmarSchwertberger:
That sounds reasonable. SearchCtrl and ComboBox inherit from wxTextEntry and therefore have a method `MSWShouldPreProcessMessage` checking for the special keys:
https://docs.wxwidgets.org/3.0/classwx_search_ctrl.html
https://docs.wxwidgets.org/3.0/classwx_combo_box.html

SpinCtrl on the other hand does not, it just uses/creates a Windows text ctrl internally.
https://docs.wxwidgets.org/3.0/classwx_spin_ctrl.html

As workaround, you could use the method outlined in the discuss thread.
(Unfortunately, it does not help to add `MSWShouldPreProcessMessage` on the Python side.)

PR #656, opened 2022-05-03
Refactored Select Tool

gentlegiantJGC:
This was originally a partial fix for #407.
Upon further investigation it was found that it was actually a bug in wxWidgets which has been fixed.

This pull request includes the refactoring from trying to fix that issue.


gentlegiantJGC:
I am going to close this because it is a bug with wxPython.
I think suppressing the events is not the correct solution and I can't see a proper workaround.

Podshot (approved):
Looks all good to me!

PR #678, opened 2022-05-14
Improved the OpenGL handling

gentlegiantJGC:
Reading into OpenGL I found some issues with how we were doing things.
OpenGL functions cannot be called until the window is shown which might be the reason some users are getting errors.
Created a function that runs the first time the window is shown. Moved all the initial OpenGL state setting into this function. Enable is run once this function is done.
Other OpenGL initialisation is moved into the drawing functions which are run after the window is shown for the first time.

### DIFF
--- a/amulet_map_editor/api/opengl/canvas/canvas.py
+++ b/amulet_map_editor/api/opengl/canvas/canvas.py
@@ -1,9 +1,11 @@
+from typing import Optional
+import logging
+
 import wx
 from wx import glcanvas
 from OpenGL.GL import (
     GL_DEPTH_TEST,
     glEnable,
-    glGenTextures,
     GL_CULL_FACE,
     glDepthFunc,
     GL_LEQUAL,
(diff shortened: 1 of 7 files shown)


frazor:
I tried building this PR, but didn't figure out how to do it.
```
    import wx
ModuleNotFoundError: No module named 'wx'
```
It's really strange, pip list gives me a version of it.
``` wxPython                  4.1.1```
Ideas?

Podshot (approved):
Looks all good to me!

Podshot:
> I tried building this PR, but couldn't figure out how to do it.
> 
> ```
>     import wx
> ModuleNotFoundError: No module named 'wx'
> ```
> 
> It's really strange, pip list gives me a version of it. ` wxPython 4.1.1` Ideas?

If you're using a virtual environment, make sure the pip you're running is the one located in the virtual environment. Sometimes the `pip` command will still use the system/global python installation even when a virtual environment is active, so try `python -m pip list` and see if wxPython is still in the output list

PR #689, opened 2022-05-19
Fixed glGenTextures error

gentlegiantJGC:
The texture was sometimes being generated in a different thread (the one that generated the chunk) leading to an error.
Moved the texture acquisition step into the draw function so it will always be called from the main thread.
Moved the resource pack class into TriMesh

This was an issue created by #678


Podshot (approved):
Looks all good to me!

PR #725, opened 2022-06-26
Allow for opening maps from the command line

zurgeg:
This PR adds support for
- Drag-n-dropping worlds onto Amulet
- Opening worlds from the CLI (i.e., allowing apps like [MultiMC](https://github.com/MultiMC/Launcher/issues/3713) to open worlds automatically)

Todo:
- [x] Test latest commit
- [x] Check for accidentally committed files 
- [x] Run black

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,21 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+wxpython = "~=4.1"
+numpy = "~=1.17"
+pyopengl = "~=3.0"
+amulet-core = "~=1.7"
+amulet-nbt = "~=1.1"
+pymctranslate = "~=1.1"
+minecraft-resource-pack = "~=1.2"
+
+[dev-packages]
+
+[requires]
+python_version = "3.9"
+
+[scripts]
+amulet = "python -m amulet_map_editor"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,196 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "6a5a8631e8df63e1f35803c0990af54874d333b7081bf9fc5b659b5452b99dd0"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.9"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": true
+            }
+        ]
+    },
+    "default": {
+        "amulet-core": {
+            "hashes": [
+                "sha256:08770c28940b8cbeebed6ae5603866f373bc4f658e307b23c701835dbb106e51",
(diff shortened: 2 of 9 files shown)


gentlegiantJGC:
I would like to support opening worlds from the command line.
This would be better implemented in the `__main__.py` module as an `else` block at the end and using argparse.
Perhaps the flag `--open-level path`. I am hesitant to use positional arguments in case our needs change.

https://github.com/Amulet-Team/Amulet-Map-Editor/blob/0.9/amulet_map_editor/__main__.py#L36-L43

```py
        try:
            app = AmuletApp(0)
            app.MainLoop()
        except Exception as e:
            log.critical(
                f"Amulet Crashed. Sorry about that. Please report it to a developer if you think this is an issue. \n{traceback.format_exc()}"
            )
            input("Press ENTER to continue.")
        else:
            # run argparse
``` 

In 0.10 I have added a method to the app instance to open and close levels which is what you would need to call.
<https://github.com/Amulet-Team/Amulet-Map-Editor/blob/0.10/amulet_map_editor/api/framework/app.py#L27-L34>

Is there need for the Pipfile? I have updated the packaging system to use the pyproject.toml and setup.cfg workflow and removed the requirements.txt file as this was quite outdated. Is there anything that would need a Pipfile that couldn't work through the pyproject.toml file?

I am going to close this pull request because the changes I have requested are effectively a full rewrite. If you want to make the changed please open another pull request.

zurgeg:
> I would like to support opening worlds from the command line. This would be better implemented in the `__main__.py` module as an `else` block at the end and using argparse. Perhaps the flag `--open-level path`. I am hesitant to use positional arguments in case our needs change.
> 
> https://github.com/Amulet-Team/Amulet-Map-Editor/blob/0.9/amulet_map_editor/__main__.py#L36-L43
> 
> ```python
>         try:
>             app = AmuletApp(0)
>             app.MainLoop()
>         except Exception as e:
>             log.critical(
>                 f"Amulet Crashed. Sorry about that. Please report it to a developer if you think this is an issue. \n{traceback.format_exc()}"
>             )
>             input("Press ENTER to continue.")
>         else:
>             # run argparse
> ```
> 
> In 0.10 I have added a method to the app instance to open and close levels which is what you would need to call. https://github.com/Amulet-Team/Amulet-Map-Editor/blob/0.10/amulet_map_editor/api/framework/app.py#L27-L34
> 
> Is there need for the Pipfile? I have updated the packaging system to use the pyproject.toml and setup.cfg workflow and removed the requirements.txt file as this was quite outdated. Is there anything that would need a Pipfile that couldn't work through the pyproject.toml file?
> 
> I am going to close this pull request because the changes I have requested are effectively a full rewrite. If you want to make the changed please open another pull request.

the pipfile was an accident :P
The reason I chose to use positional arguments was to make MMC support easier without needing much modification. On top of that, unless we use positional arguments, we can't have drag-n-drop support, but feel free to make your own decisions regarding this. Thanks for taking a look at this.

PR #729, opened 2022-06-30
Updated _create_directory_picker with new option

StealthyExpertX:
The directory method now has an option for setting the default path if supplied, if none are supplied will be set to a blank string which was the previous behavior.

Code by gentlegiantJGC, implemented by StealthyX

- Sorry, missed this in the last pull request.


StealthyExpertX:
Not sure what I'm doing wrong with black? I installed it via pip and ran it on the file before uploading to github.

gentlegiantJGC:
Make sure the version of black you are running is the newest version. Sometimes different versions of black can format the code differently.
`python -m black .` run from the root directory should format all files in the directory.

StealthyExpertX:
Yeah, I just updated my black installation to a newer version so hopefully, that resolves it next time.

PR #735, opened 2022-07-05
Added new UI button option with examples.

StealthyExpertX:
Added a new UI button option with an example, I also have done the due diligence to update the examples in 3_fixed_function_pipeline.py

Added a new private function _create_button() that adds a new button UI type for the simple plugin UI API.
The first argument is the "button name", and the second argument is the function to call as an event when the button is clicked/pressed by the user.

#custom function you want to run from the button click event.
def function_name():
    print("you have clicked a button.")

#button input type with two arguments.
operation_options = {
    "Button input": ["button", "button name", function_name]
}

### DIFF
--- a/amulet_map_editor/programs/edit/api/operations/ui/fixed_pipeline.py
+++ b/amulet_map_editor/programs/edit/api/operations/ui/fixed_pipeline.py
@@ -207,21 +207,17 @@ class FixedFunctionUI(wx.Panel, DefaultOperationUI):
         self._options[option_name] = option
 
     def _create_button(self, option_name: str, options: Sequence):
-        def _get_on_click(self, callback):
-            def _on_click(event):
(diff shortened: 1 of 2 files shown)


gentlegiantJGC:
A couple of issues.
The `callback` variable is shadowing the outer scope. Try to avoid name shadowing to avoid confusion.
A options value of `[]` will lead to callback being undefined at line 224 and an options value of `["name"]` will lead to an unpacking error on line 217.
Is there benefit to having a button with no callback function?

I think it would make sense to handle unpacking the variables in the actual function definition rather than within the function which will make this a bit simpler.
I will make these changes.

gentlegiantJGC (approved):

StealthyExpertX:
Alright, thanks for the pointers!

PR #836, opened 2022-11-23
improve language zh_CN

1834423612:
> Add and Change zh-CN translations

Hi, 
I have made some modifications to the translation of zh_CN according to the operation interface and usage functions and added a few places that have not been translated.

Since the translation from the original language (English) to Chinese may have some doubts for users, I explained the Amulet interface according to the method that Chinese users can easily understand waht does it means. 
And I added the explanation in parentheses after the translated content.

This is my first time to Pull requests, something maybe was not very well.😊 
Thank you.

### DIFF
--- a/amulet_map_editor/lang/zh_CN.lang
+++ b/amulet_map_editor/lang/zh_CN.lang
@@ -109,7 +109,6 @@ program_3d_edit.tab_name=3D 编辑器
 
 ## Canvas
 program_3d_edit.canvas.please_wait=渲染器正在初始化, 请稍作等待
-program_3d_edit.canvas.downloading_bedrock_vanilla_resource_pack=正在下载 Bedrock vanilla 纯净资源包
 program_3d_edit.canvas.downloading_bedrock_vanilla_resource_pack=正在下载基岩版纯净资源包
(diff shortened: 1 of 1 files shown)


gentlegiantJGC (approved):
Thanks. This looks good

amulet_map_editor/lang/zh_CN.lang
```diff
@@ -98,6 +109,7 @@ program_3d_edit.tab_name=3D 编辑器
 
 ## Canvas
 program_3d_edit.canvas.please_wait=渲染器正在初始化, 请稍作等待
+program_3d_edit.canvas.downloading_bedrock_vanilla_resource_pack=正在下载 Bedrock vanilla 纯净资源包
 program_3d_edit.canvas.downloading_bedrock_vanilla_resource_pack=正在下载基岩版纯净资源包
```
gentlegiantJGC: These two lines are duplicated. Please delete one of them.

gentlegiantJGC (changes requested):

amulet_map_editor/lang/zh_CN.lang
1834423612: Sorry, I forget it, now is changed😃

gentlegiantJGC (approved):

PR #984, opened 2023-09-28
Console world converter

yri066:
Added the use of console arguments, to convert the world via console


gentlegiantJGC:
Thanks for the contribution. We currently aren't adding new features to this editor while we work on the Qt port.
It would be nice to support something like this in the new version but we will need to have a think how is best to structure console arguments so that we are not locked into only using the CLI for conversion.

PR #1174, opened 2025-06-17
Use a 'default' instead of a 'decorative' font family

### LINKED ISSUE #1172: [Feature Request] Consider using a different font family for the menu
### The Problem
Currently, a "decorative" font family is used in some places:

https://github.com/Amulet-Team/Amulet-Map-Editor/blob/679d42367da5ca3516bde9cc3c56a3ef9c45405b/amulet_map_editor/api/framework/pages/main_menu.py#L39
https://github.com/Amulet-Team/Amulet-Map-Editor/blob/679d42367da5ca3516bde9cc3c56a3ef9c45405b/amulet_map_editor/api/framework/pages/main_menu.py#L44

This might default to a very unsuitable font like Impact, if installed:
Image

### Feature Description
Consider using a different [fount family](https://docs.wxpython.org/wx.FontFamily.enumeration.html) (like `wx.FONTFAMILY_DEFAULT`) with a thicker [weight](https://docs.wxpython.org/wx.FontWeight.enumeration.html) (i.e., `wx.FONTWEIGHT_BOLD`) instead, or provide a custom list of fonts to be preferred, similar to the [CSS system font stack](https://css-tricks.com/snippets/css/system-font-stack/).

### Alternatives
The only alternative I can think of would be for the user to configure the system fonts differently, so that Impact is no longer prioritized. But this will also affect other applications and not every user might know how to do it.

gwuen:
On Unix systems, the wxWidgets / wxPython font family `wx.DECORATIVE` defaults to Impact [^1]. For a more suitable UI font, this is changed to `wx.DEFAULT`.

On modern macOS and Windows systems, this is expected to have minimal impact (no pun intended), as `wx.DECORATIVE` already seems to be equivalent to `wx.DEFAULT` on both platforms [^2][^3].

Even with its original intent, the 'decorative' font family is not a good match for the UI, as it corresponds to the `fantasy` font family in CSS, that mostly includes gimmicky fonts like Papyrus or Jokerman [^4].

| Before (Impact) |
|:--:|
| Before |

| After (Noto Sans) |
|:--:|
| After |

Resolves #1172

[^1]: https://github.com/wxWidgets/wxWidgets/blob/18da5b8269706472f6c0fa09dcbae605a477da19/src/unix/fontutil.cpp#L274-L277
[^2]: https://github.com/wxWidgets/wxWidgets/blob/18da5b8269706472f6c0fa09dcbae605a477da19/src/osx/carbon/font.cpp#L218-L220
[^3]: https://github.com/wxWidgets/wxWidgets/issues/24178
[^4]: https://web.archive.org/web/20130126125901/http://www.codestyle.org/css/font-family/sampler-Fantasy.shtml

### DIFF
--- a/amulet_map_editor/api/framework/pages/main_menu.py
+++ b/amulet_map_editor/api/framework/pages/main_menu.py
@@ -36,12 +36,12 @@ class AmuletMainMenu(wx.Panel, BasePageUI):
         name_sizer.Add(icon, flag=wx.CENTER)
 
         self._amulet_name = wx.StaticText(self)
-        self._amulet_name.SetFont(wx.Font(40, wx.DECORATIVE, wx.NORMAL, wx.NORMAL))
+        self._amulet_name.SetFont(wx.Font(40, wx.DEFAULT, wx.NORMAL, wx.NORMAL))
         name_sizer.Add(
             self._amulet_name, flag=wx.CENTER | wx.LEFT | wx.RIGHT, border=10
         )
         name_sizer.Add(icon2, flag=wx.CENTER)
-        button_font = wx.Font(20, wx.DECORATIVE, wx.NORMAL, wx.NORMAL)
+        button_font = wx.Font(20, wx.DEFAULT, wx.NORMAL, wx.NORMAL)
         self._open_world_button = wx.Button(self, size=(400, 70))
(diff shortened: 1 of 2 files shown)


gentlegiantJGC (approved):
Looks good

PR #1184, opened 2025-07-29
Create docker-image.yml

dragonruler1000:
Added a workflow to build the docker image automatically.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -3,7 +3,7 @@ name: Docker Image CI
 
 on:
   workflow_run:
-    workflows: [Python-build]
+    workflows: [python-build]
     types:
       - completed
 
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -2,8 +2,10 @@ name: Docker Image CI
 
 
 on:
-  release:
-    types: [published]
+  workflow_run:
+    workflows: [First Continuous Deployment]
+    types:
+      - completed
 
 jobs:
   build:
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -20,10 +20,9 @@ jobs:
     - name: Save Docker image to tar archive
       run: |
         docker save amulet-editor:latest -o amulet-editor.tar
-
-    - name: Upload image to GitHub Release
-      uses: softprops/action-gh-release@v1
-      with:
-        files: amulet-editor.tar
+        
+    - name: Upload Release Asset
       env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        GITHUB_TOKEN: ${{ github.TOKEN }}
+      run: |
+        gh release upload ${{github.event.release.tag_name}} ./dist/${{env.BUILT_PATH}}#${{env.BUILT_PATH}}
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,9 +1,9 @@
 name: Docker Image CI
 
+
 on:
-  push:
-    tags:
-      - "v*"  # Trigger on version tags like v1.0, v0.10, etc.
+  release:
+    types: [published]
 
 jobs:
   build:
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -25,4 +25,4 @@ jobs:
       env:
         GITHUB_TOKEN: ${{ github.TOKEN }}
       run: |
-        gh release upload ${{github.event.release.tag_name}} ./dist/${{env.BUILT_PATH}}#${{env.BUILT_PATH}}
+        gh release upload ${{github.event.release.tag_name}} amulet-editor.tar#Amulet-v${{ github.event.release.tag_name }}-docker.tar
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,29 @@
+name: Docker Image CI
+
+on:
+  push:
+    tags:
+      - "v*"  # Trigger on version tags like v1.0, v0.10, etc.
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v4
+
+    - name: Build Docker image
+      run: |
+        docker build . --file Dockerfile --tag amulet-editor:latest
+
+    - name: Save Docker image to tar archive
+      run: |
+        docker save amulet-editor:latest -o amulet-editor.tar
+
+    - name: Upload image to GitHub Release
+      uses: softprops/action-gh-release@v1
+      with:
+        files: amulet-editor.tar
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,6 +1,5 @@
 name: Docker Image CI
 
-
 on:
   workflow_run:
     workflows: [python-build]
@@ -17,14 +16,19 @@ jobs:
 
     - name: Build Docker image
       run: |
-        docker build . --file Dockerfile --tag amulet-editor:latest
+        docker build \
+          . \
+          --file Dockerfile \
+          --tag amulet-editor:latest \
+          --build-arg AMULET_VERSION=${{ github.event.release.tag_name }}
 
     - name: Save Docker image to tar archive
       run: |
         docker save amulet-editor:latest -o amulet-editor.tar
-        
+
     - name: Upload Release Asset
       env:
-        GITHUB_TOKEN: ${{ github.TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        gh release upload ${{github.event.release.tag_name}} amulet-editor.tar#Amulet-v${{ github.event.release.tag_name }}-docker.tar
+        gh release upload ${{ github.event.release.tag_name }} \
+          amulet-editor.tar#Amulet-v${{ github.event.release.tag_name }}-docker.tar
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -3,7 +3,7 @@ name: Docker Image CI
 
 on:
   workflow_run:
-    workflows: [First Continuous Deployment]
+    workflows: [Python-build]
     types:
       - completed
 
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
+name: Docker Image CI
+
+on:
+  push:
+    tags:
+      - "v*"  # Trigger on version tags like v1.0, v0.10, etc.
```
gentlegiantJGC: We don't add a `v` prefix to version numbers.
See here for how to implement this
https://github.com/Amulet-Team/Amulet-Map-Editor/blob/4889915b0f7a30416b9994929241dc657f6848b4/.github/workflows/python-build.yml#L6C1-L8C23

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Save Docker image to tar archive
+      run: |
+        docker save amulet-editor:latest -o amulet-editor.tar
+
+    - name: Upload image to GitHub Release
+      uses: softprops/action-gh-release@v1
+      with:
+        files: amulet-editor.tar
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
gentlegiantJGC, line 31: This can be implemented with this github command
https://github.com/Amulet-Team/Amulet-Map-Editor/blob/4889915b0f7a30416b9994929241dc657f6848b4/.github/workflows/python-build.yml#L66C1-L69C108
Docs
https://cli.github.com/manual/gh_release_upload

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.9-bullseye
+RUN apt-get update && apt install libgtk-3-dev dbus-x11 wget libnotify4 -y
+RUN pip install --upgrade pip
+RUN wget https://extras.wxpython.org/wxPython4/extras/linux/gtk3/debian-11/wxPython-4.2.0-cp39-cp39-linux_x86_64.whl && pip install wxPython-4.2.0-cp39-cp39-linux_x86_64.whl
+RUN pip install amulet-map-editor==0.10.44
```
gentlegiantJGC: I would prefer if this was not hard coded.
You have already cloned the repository so can you do `pip install .` to install from the current directory?
If docker changes the working directory you may need to change the path.

gentlegiantJGC (changes requested):

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.9-bullseye
+RUN apt-get update && apt install libgtk-3-dev dbus-x11 wget libnotify4 -y
+RUN pip install --upgrade pip
+RUN wget https://extras.wxpython.org/wxPython4/extras/linux/gtk3/debian-11/wxPython-4.2.0-cp39-cp39-linux_x86_64.whl && pip install wxPython-4.2.0-cp39-cp39-linux_x86_64.whl
```
gentlegiantJGC, line 4: Where are you requesting the version of Python?
Is 3.9 the default for the runner?
You can use `actions/setup-python@v5` to ensure a specific version of python is installed
https://github.com/Amulet-Team/Amulet-Map-Editor/blob/4889915b0f7a30416b9994929241dc657f6848b4/.github/workflows/python-build.yml#L25C1-L29C53

Dockerfile
dragonruler1000: I didn't realise that it was hard coded. I think that was just because i had copied the info from pypi.  updated it in the docker file on my fork so it should update in the PR.

Dockerfile
dragonruler1000: The verison of python is set in the FROM line so yes python 3.9 is the version of python that it is using.

.github/workflows/docker-image.yml
dragonruler1000: updated on my branch.

dragonruler1000 (commented):
Fixed/implemented all of your issues and suggestions.

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    - name: Save Docker image to tar archive
+      run: |
+        docker save amulet-editor:latest -o amulet-editor.tar
+        
+    - name: Upload Release Asset
+      env:
+        GITHUB_TOKEN: ${{ github.TOKEN }}
+      run: |
+        gh release upload ${{github.event.release.tag_name}} ./dist/${{env.BUILT_PATH}}#${{env.BUILT_PATH}}
```
gentlegiantJGC: `./dist/${{env.BUILT_PATH}}#${{env.BUILT_PATH}}`
This path is incorrect.
The part before the `#` is the path to upload and the part after is the file name the user will see when downloading.
`env.BUILT_PATH` was a variable set in the workflow you copied it from.

I suggest something like this `amulet-editor.tar#Amulet-v${{ github.event.release.tag_name }}-os-architecture.tar`
Swap out `os-architecture` for sensible values so the user knows which version to download.

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+RUN pip install amulet-map-editor
```
gentlegiantJGC: This does not solve the original issue. The new version may not have been uploaded to pypi when this command is run.
`pip install .` will install it from the current directory.

gentlegiantJGC (changes requested):

Dockerfile
dragonruler1000: Fixed this by having the workflow run after python-build is completed so it is will have been pushed to pypi.

dragonruler1000 (commented):
That should fix all the issues.

Dockerfile
gentlegiantJGC: why not use the command I sent?

dragonruler1000: because im also wanting the docker file to be able to be built by the user after they download it. because when you run the docker image you are running the Dockerfile and if people want to update the docker image they can just run docker build path/to/Dockerfile and i don't feel like having the Dockerfile clone the git hub repo is going to be a good way to have it build it.

dragonruler1000:
Just did some commits that will make it so Docker will build the latest version of amulet and not use the cached version every time.

Dockerfile
gentlegiantJGC: Okay. I have no experience with docker

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.9-bullseye
+RUN apt-get update && apt install libgtk-3-dev dbus-x11 wget libnotify4 -y
+RUN pip install --upgrade pip
+RUN wget https://extras.wxpython.org/wxPython4/extras/linux/gtk3/debian-11/wxPython-4.2.0-cp39-cp39-linux_x86_64.whl && pip install wxPython-4.2.0-cp39-cp39-linux_x86_64.whl
+ARG AMULET_VERSION=0.10.44  # defaults to version 0.10.44 if no arg is given in build command.
+RUN pip install amulet-map-editor==$AMULET_VERSION
```
gentlegiantJGC, line 6: I like the variable configuration but I don't like the hard coded default value. I would prefer to set this and forget about it. I would also like to support alpha and beta releases.

Does docker have an if else statement?
Something like
```
ARG AMULET_VERSION=RELEASE
IF AMULET_VERSION==RELEASE {
    RUN pip install --upgrade --upgrade-strategy eager amulet-map-editor
} ELIF AMULET_VERSION==BETA {
    RUN pip install --upgrade --upgrade-strategy eager amulet-map-editor>=0b0
} ELIF AMULET_VERSION==ALPHA {
    RUN pip install --upgrade --upgrade-strategy eager amulet-map-editor>=0a0
} ELSE {
    RUN pip install amulet-map-editor==$AMULET_VERSION
}
```

I think that should work

gentlegiantJGC (changes requested):

Dockerfile
dragonruler1000: I'll probably just leave it blank. Yes, there is a way to have an if-else statement but it is more complicated than I want to try. I don't think most people are going to care if they can just install the beta version by passing BETA into the argument. If someone wants to do that they can add the functionality themselves.

dragonruler1000: Or if passing obo/oba doesn't work via the command line you can always just manually edit the Docker file to get it to work.

gentlegiantJGC (approved):
Looks good. I will do an alpha release to test it.

gentlegiantJGC:
Here is the release. If everything goes well the file should appear here
https://github.com/Amulet-Team/Amulet-Map-Editor/releases/tag/0.10.45a0

This is the main build
https://github.com/Amulet-Team/Amulet-Map-Editor/actions/runs/16726368107

Your new workflow should appear here when that completes
https://github.com/Amulet-Team/Amulet-Map-Editor/actions/workflows/docker-image.yml

gentlegiantJGC:
@dragonruler1000 
```
#9 [5/6] RUN pip install amulet-map-editor==
#9 0.389 ERROR: Invalid requirement: 'amulet-map-editor==': Expected end or semicolon (after name and no valid version specifier)
#9 0.389     amulet-map-editor==
#9 0.389                      ^
#9 ERROR: process "/bin/sh -c pip install amulet-map-editor==$AMULET_VERSION" did not complete successfully: exit code: 1
------
 > [5/6] RUN pip install amulet-map-editor==:
0.389 ERROR: Invalid requirement: 'amulet-map-editor==': Expected end or semicolon (after name and no valid version specifier)
0.389     amulet-map-editor==
0.389                      ^
------
Dockerfile:6
--------------------
   4 |     RUN wget https://extras.wxpython.org/wxPython4/extras/linux/gtk3/debian-11/wxPython-4.2.0-cp39-cp39-linux_x86_64.whl && pip install wxPython-4.2.0-cp39-cp39-linux_x86_64.whl
   5 |     ARG AMULET_VERSION  # defaults to the latest version if no arg is given in build command.
   6 | >>> RUN pip install amulet-map-editor==$AMULET_VERSION
   7 |     RUN apt-get install libSDL2-2.0 -y
   8 |     
--------------------
ERROR: failed to build: failed to solve: process "/bin/sh -c pip install amulet-map-editor==$AMULET_VERSION" did not complete successfully: exit code: 1
```

gentlegiantJGC:
I think the issue is that the workflow does not have access to `github.event.release.tag_name` because it isn't run from a publish event.

gentlegiantJGC:
Here is the full output.
[job-logs.txt](https://github.com/user-attachments/files/21580528/job-logs.txt)

dragonruler1000:
I know im trying to fix it right now. I'm going to open a new pull request in a little bit that will hopefully fix the problem.

dragonruler1000:
@gentlegiantJGC should be fixed by #1187

PR #1185, opened 2025-08-04
Fix workflow name from #1184

gentlegiantJGC:


dragonruler1000:
I guess i thought that the workflow file was the same as the work flow name. I probably should have checked that.

PR #1186, opened 2025-08-04
Fix docker workflow

gentlegiantJGC:
Support alpha and beta releases


dragonruler1000:
#1187 should add support for pre-releases without needing the if statements.

gentlegiantJGC:
@dragonruler1000 I think this should solve all the issues.

gentlegiantJGC:
Okay that failed. Docker must move the CWD somewhere else.

PR #1187, opened 2025-08-04
Hopfully fix the issue from before

dragonruler1000:
Updated the docker image workflow to use a GitHub repo var. it should work after adding `AMULET_VERSION` to the Repo vars.


gentlegiantJGC:
I don't like that solution. I have an idea I will implement in my pull request.

PR #1193, opened 2025-08-27
Added an experimental touch-screen mode

Max-RM:
An experimental touch-screen mode has been added for touchscreen devices, such as x86 Windows tablets, tablets from the Microsoft Surface line, other touchscreen devices and phones with unofficially installed Windows ARM (Renegade Project) which was my case.
image_2025-08-26_18-40-55

The sensitivity of the camera can now take fractional, not just integer values, as it is necessary for touch control mode (recommended to use 0.2), in Options > Controls... A new switch "Enable touchscreen mode (experimental)" has been added, which enables touch control mode, which is disabled by default.
When turned on, it adds 6 buttons to control the camera's position in space (forward, backward, left, right, up, down), they work as switches, after pressing it will be active and turned off after pressing it again, you can also simulate diagonal movement by activating, for example, the forward and right buttons.
When the touch mode is activated up, two new buttons appear: "Touch Controls" and "Camera/Selector". The "Touch Controls" button can turn on and off 6 buttons for controlling the camera's position in space, this is necessary, since when using standard operations and third-party plug-ins, these buttons can overlap part of the interface, while this button allows you to temporarily disable them.
The Camera / Selector button is needed to switch the rotation modes of the camera's view direction (camera) and the selection mode (Selector), since there are big problems using the right and left mouse clicks on the touchscreen, this button helps to switch between modes so that it is easy.
While using the "Camera" mode, any interaction with the selection areas is blocked. When the "Selector" mode is turned on, to select the area correctly, you first need to long tap the desired block, then stretch from this block to another desired point, then if you need to stretch the area in one direction, you can not just poke along the edge of the selection area and stretch this, but first you need to long tap on the desired face, and as soon as it turns blue, you can stretch it in the right direction. If a mistake was made during the selection, you can simply cancel the action via the button at the top. 

This is the most complete implementation of touch-screen control that could be done without having to use a mouse through an adapter.


gentlegiantJGC:
This needs to be based on the `0.10` branch

PR #1195, opened 2025-08-29
Update docker-image.yml

dragonruler1000:
Making it so the Dockerfile is actually compressed not just put in a archive.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -29,4 +29,4 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
         gh release upload ${{ github.event.release.tag_name }} \
-          amulet-editor.tar.gz#Amulet-v${{ github.event.release.tag_name }}-docker.tar.xz
+          amulet-editor.tar.xz#Amulet-v${{ github.event.release.tag_name }}-docker.tar.gz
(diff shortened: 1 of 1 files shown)


gentlegiantJGC:
Is the mixing of `.xz` and `.gz` intentional?

dragonruler1000:
> Is the mixing of `.xz` and `.gz` intentional?

Yes, `.xz` does higher compression it is ~50-60% smaller then the original file while `.gz` is only ~30-40% smaller.

gentlegiantJGC:
Just to clarify, your first command creates a `.tar.xz` file and the second command tries uploading a `.tar.gz` file.
Won't this break?

dragonruler1000:
Didn't realize that i had swapped the 2. Should be fixed now.

gentlegiantJGC:
@dragonruler1000 are you able to update this to use Python 3.10? I tried doing it myself but wxPython does not have builds for 3.10 for the OS you are using.

dragonruler1000:
> @dragonruler1000 are you able to update this to use Python 3.10? I tried doing it myself but wxPython does not have builds for 3.10 for the OS you are using.

I probably can by switching to Ubuntu I'll try to do that later today. As long as the latest version of wxPython is for Python 3.10 it should work.

gentlegiantJGC:
Thank you. Do you have PayPal or some other way to donate money to you?

dragonruler1000:
i do it is paypal.me/ZachariahJackson41 plus i also have other methods that are listed in my emberborn repo. i will get the updated docker file pushed after i figure out what version of wxpython i need to use.

PR #1246, opened 2026-01-07
Update Dockerfile

dragonruler1000:
update to run on python 3.10

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,33 @@
-FROM python:3.9-bullseye
-RUN apt-get update && apt install libgtk-3-dev dbus-x11 wget libnotify4 -y
-RUN pip install --upgrade pip
-RUN wget https://extras.wxpython.org/wxPython4/extras/linux/gtk3/debian-11/wxPython-4.2.0-cp39-cp39-linux_x86_64.whl && pip install wxPython-4.2.0-cp39-cp39-linux_x86_64.whl
+FROM ubuntu:22.04
+
+# Prevent interactive prompts during apt install
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install Python 3.10 (default on Ubuntu 22.04) and required dependencies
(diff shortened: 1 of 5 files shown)


gentlegiantJGC:
Is this ready to review?

dragonruler1000:
it is ready. it also didn't take to much to do.

Dockerfile
```diff
@@ -1,19 +1,34 @@
-FROM python:3.9-bullseye
-RUN apt-get update && apt install libgtk-3-dev dbus-x11 wget libnotify4 -y
-RUN pip install --upgrade pip
-RUN wget https://extras.wxpython.org/wxPython4/extras/linux/gtk3/debian-11/wxPython-4.2.0-cp39-cp39-linux_x86_64.whl && pip install wxPython-4.2.0-cp39-cp39-linux_x86_64.whl
+FROM ubuntu:22.04
+
+# Prevent interactive prompts during apt install
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install Python 3.10 (default on Ubuntu 22.04) and required dependencies
```
gentlegiantJGC, line 6: Can you hard code the python version?
Our runner is using `ubuntu-latest` which at the time of writing is `24.04`
If not change the runner to use `ubuntu-22.04`

gentlegiantJGC (changes requested):

Dockerfile
dragonruler1000: the version of ubuntu that the runner is using doesn't mater for the docker file as that is just what it is being built on and docker is close to just running a script or app in side a dedicated vm without the overhead of setting up a vm. i have set the docker image to spesificly install python3.10 but just installing python3 works as well.

dragonruler1000 (commented):
updated to install python3.10 specifically.

gentlegiantJGC (approved):

gentlegiantJGC:
I will do a test build in the next couple of days

PR #1270, opened 2026-02-04
Add Pocket Edition path on MacOS

### LINKED ISSUE #1269: There's no MCEE world's page in the Amulet on MacOS.
Hi.
Recently I bought old iMac 2012, and found that Amulet Map Editor have no standart path for MCEE world's on MacOS.
Here it is:
/Users/USERNAME/Library/Application Support/minecraftpe/games/com.mojang/minecraftWorlds/

That's for old MCEE world storage paths, in 1.21.X versions it may be changed but I can't test it because it crashes for me for no reason, probably because of OS version 10.15 (although it should be compatible).

Maybe it's will be good to add it?

Image

gentlegiantJGC:
Fixes #1269

### DIFF
--- a/amulet_map_editor/api/wx/ui/select_world.py
+++ b/amulet_map_editor/api/wx/ui/select_world.py
@@ -136,6 +136,20 @@ elif platform == "darwin":
             ),
         )
     )
+    minecraft_world_paths.append(
+        (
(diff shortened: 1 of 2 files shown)


Max-RM:
IMG_20260204_211435_326
Yeah, it works now.
Thanks!

PR #1302, opened 2026-02-28
Use window manager from OS for control dialog

jtmoree-github-com:
The controls dialog does not have standard OS buttons for the window.   Maximizing and Closing are manual operations and not fast like other applications.  (yes closing is available with the buttons on the bottom but not fast).  Escape doesn't do anything like a regular dialog so let's treat this as a real window.

This patch allows the OS to control the window and add standard buttons.  Only tested on Windows so far.

### DIFF
--- a/amulet_map_editor/api/wx/util/key_config.py
+++ b/amulet_map_editor/api/wx/util/key_config.py
@@ -358,17 +358,13 @@ class KeyConfigDialog(SimpleDialog):
         entries: Sequence[KeyActionType],
         fixed_keybinds: KeybindContainer,
         user_keybinds: KeybindContainer,
+        style: int = wx.CAPTION
+        | wx.CLOSE_BOX
+        | wx.MAXIMIZE_BOX
+        | wx.SYSTEM_MENU
+        | wx.RESIZE_BORDER,
     ):
-        super().__init__(
-            parent,
-            lang.get("key_config.key_select"),
-            style=wx.CAPTION
-            | wx.CLOSE_BOX
-            | wx.MAXIMIZE_BOX
-            | wx.MINIMIZE_BOX
-            | wx.SYSTEM_MENU
-            | wx.RESIZE_BORDER,
-        )
+        super().__init__(parent, lang.get("key_config.key_select"), style=style)
 
         self._key_config = KeyConfig(
             self, selected_group, entries, fixed_keybinds, user_keybinds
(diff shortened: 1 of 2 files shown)


gentlegiantJGC (changes requested):
I like this suggestion.
It can be more simpliy implemented by modifying the `SimpleDialog` class to take a style argument.
The dialog is being shown modally so is minimise useful?
What does `SYSTEM_MENU` do?

jtmoree-github-com:
I found that as I added more hotkeys I always wanted to maximize the keyboard controls mapping window.  Since there were no OS controls that was not possible (or I didn't know how).

amulet_map_editor/api/wx/util/key_config.py
```diff
@@ -359,7 +359,28 @@ def __init__(
...
+        # Initialize SimpleDialog but override the style to include window management
+        wx.Dialog.__init__(
+            self,
+            parent,
+            title=lang.get("key_config.key_select"),
+            style=wx.CAPTION
+            | wx.CLOSE_BOX
+            | wx.MAXIMIZE_BOX
+            | wx.MINIMIZE_BOX
+            | wx.SYSTEM_MENU
```
jtmoree-github-com: Im not sure why SYSTEM_MENU is here.  maybe leftover from some experimental code?   Guess I should remove it....

jtmoree-github-com: oh.  i wonder if it's related to the window manager putting the system menu in the top left corner on windows.  it's the Icon that has close and other stuff in it.

jtmoree-github-com: yes.  that's it.

jtmoree-github-com: i dont see a reason to remove the minimize.  if people dont use it it's not in the way.   I changed to the style argument.

gentlegiantJGC: This is only ever shown modally.
If the window is minimised the user cannot interact with the rest of the program and it is confusing to the user how to get back to the program. I would prefer if minimising was disabled.

Perhaps the style argument could be moved into the KeyConfigDialog constructor so that the caller can choose the options.

jtmoree-github-com: oh right.  minimize could be a problem.   removed.

gentlegiantJGC:
I am still confused what SYSTEM_MENU does.
This is with it enabled.
image
This is with it disabled.
image
They look the same.

gentlegiantJGC:
Perhaps it isn't used. It seems to be part of the default style so I have no issue leaving it.

gentlegiantJGC:
Looks good other than the formatting change.
I would usually fix this myself but Github is not letting me open it.

PR #1310, opened 2026-03-15
retool convert inteface

jtmoree-github-com:
Allow New world for target
default new world name to current. allow edit
Use radio for new vs existing
filter existing worlds (not current one; not current type) to only valid targets show icon for existing world to right of current
file paths as highlight/selectable text


jtmoree-github-com:
I find the current convert interface confusing.    This is intuitive as it explicitly labels what we are doing
bedrock =>java (or vice versa)
new or existing world as target

PR #1349, opened 2026-05-11
Remove invalid --upgrade flag from pip wheel commands

keenanbuckley:
The Docker build currently fails at step 8/14. `pip wheel` is invoked with `--upgrade --upgrade-strategy eager`, which appear to be `pip install` flags rather than `pip wheel` flags, and pip exits with `no such option: --upgrade`.

`pip wheel` already resolves to the newest version satisfying the requirement, so the flags should not be needed. Removing them lets the build complete.


gentlegiantJGC (approved):
Thanks

Issue #1369, opened 2026-07-07
Chunk loading is very slow on large worlds; top-down view cannot show the whole world


babagugliana:
On decent-sized worlds the editor is hard to work with: the top-down view loads chunks in a spiral from the camera and restarts the spiral every time the camera moves, chunk generation is limited to a single thread with a 10ms sleep per chunk, and the periodic garbage collector discards loaded chunks shortly after panning away. There is also no way to see the whole world at once - everything beyond render distance is empty.

I have a working branch that addresses this and would like to contribute it:

- The spiral iterator is replaced with a persistent priority queue (camera movement re-prioritises instead of restarting) and a worker pool, with a memory-budget LRU instead of the distance purge, so revisiting an area is instant.
- The top-down view renders a whole-world map (top visible block colour per column with relief shading) built by a background scanner and persisted to the app cache, so it appears instantly on later opens. Edits, undo/redo and chunk creation/deletion update the map via the level's changed-chunk tracking.
- Bedrock worlds additionally get a raw LevelDB fast path for the scan (Data3D heightmap + top-block decode only, colours resolved through the normal translation pipeline once per unique block state): measured 1105 chunks/s vs 86 chunks/s through full translation on a real 2GB 1.26 world, with per-chunk fallback so correctness never regresses.
- Top-down navigation: middle-mouse drag panning, smooth time-based movement, larger zoom range, and a Deselect All button in the select tool.

Branch: https://github.com/babagugliana/Amulet-Map-Editor/tree/improv-chunk-loading (3 logical commits, 60 new unit tests, black-clean)

I am opening a PR alongside this issue so the code is visible. Happy to split it into a series or adjust the approach to whatever fits your review process.

PR #1370, opened 2026-07-07
Faster chunk loading and a whole-world map for the top-down view

babagugliana:
Related issue: #1369

### Motivation

On large worlds the editor was painful to use: chunk loading spiralled out from the camera and restarted on every movement, ran on a single thread capped at ~100 chunks/s, and a distance-based garbage collector threw away loaded chunks shortly after panning away. Beyond render distance there was nothing to see, so navigating a big world meant constantly waiting.

### What this does

Three commits, each self-contained:

1. **Rework chunk loading** - persistent distance-priority queue (camera movement re-prioritises rather than restarting), a small worker pool (default `min(4, cpu - 2)` threads - more mostly fight the GIL) instead of one thread with a 10ms sleep, and a memory-budget LRU (default RAM/4, configurable) instead of the 10s distance purge, so revisiting an area is instant. Edit detection is event-driven instead of rescanning.

2. **Whole-world overview map** - in top-down mode the world renders as a map: top visible block colour per column with relief shading, generated by a background scanner (nearest-camera-first) and persisted as 512x512-block PNG tiles under the app cache dir (`.../AmuletMapEditor/Cache/world_overview/`), so reopening a world shows the whole map immediately. Edits, undo/redo and chunk creation/deletion update the map through `level.chunks.changed_chunks()` processed on the worker pool. Bedrock (LevelDB) worlds use a raw fast path (Data3D heightmap + only the sub-chunk records holding each column's top block; colours resolved through the normal translation pipeline once per unique block state, memoised) with per-chunk fallback to the translation path on any parse problem. Top-down defaults to map-only at every zoom; an option restores the 3D detail layer when zoomed in, or always. Java and unknown formats use the translation path throughout - behaviour there is unchanged apart from the loader rework.

3. **Top-down navigation/UX** - middle-mouse drag panning (1:1 with the cursor), time-scaled held-key movement on a 16ms input timer (removes the 33ms-step judder; effective speeds unchanged, nudge-button repeat rates preserved), the selection pointer no longer synchronously loads chunks on the UI thread while panning, zoom range extended to 5000 half-height, and a localised "Deselect all" button in the select tool.

### Measured results (real 2.1GB Bedrock 1.26 world, 16-core Windows 10)

- Overview scan: **1105 chunks/s** single-thread via the fast path vs **86 chunks/s** through full universal translation (~13x per thread; the raw LevelDB reads also release the GIL so workers scale).
- First full-world map: minutes -> well under a minute; subsequent opens: instant from cache.
- Panning no longer restarts any work; geometry within the memory budget is never rebuilt.

### New options (all with safe defaults)

Overview on/off - chunk worker threads - geometry memory budget (MB) - 3D detail in top-down (map only / when zoomed in / always) - world map rescan button with live progress.

### Notes for reviewers

- All OpenGL calls remain on the main thread; workers only produce numpy arrays. Shared mesher state is guarded by a single `model_lock`.
- Tile cache writes are atomic (temp file + `os.replace`, PNG before metadata) and crash-safe; interrupted scans resume where they stopped.
- Edits made in-session always rescan through the level API (the db only holds saved state) - the raw fast path is used only for initial sweeps.
- 60 new unit tests (116 total passing); `black .` clean.
- Happy to split this into a PR series if that fits your review process better.

### Known limitations

- The scanner keeps scanned tile pixels in RAM for the session (~1MB per 512x512-block region); fine for large worlds, could be evicted to disk for truly enormous ones.
- Legacy (pre-1.18) Bedrock chunks use the translation path (no Data2D fast path yet).

Demo: world opens with the full map from cache, middle-mouse panning

amulet_fast


gentlegiantJGC:
I think this is a cool idea.
The current core library is not thread safe which is why only one thread is used. This is something I have been working on in the rewrite.
There are also a lot of pointless changes in this pull request that make me believe it was AI generated.

PR #1392, opened 2026-09-19
V0.10.64b0_i18n_Fix

BoringMan314:
## Complete remaining i18n, Java 26.3 save compatibility, and recent-world list

- **Fix**: Wire leftover hardcoded English to lang keys, including Options, toolbar (Select, Paste, Operation, Import, Export, Chunk), run/export/save/close-world progress, unsaved-changes dialogs, import/export, copy/paste, native language names, OK/Cancel, operation names (Clone, Fill, etc.), and error/plugin-folder dialogs. Dialog titles no longer fall back to English Message (closing a world uses Unsaved Changes).
- **Add**: Java 26.3 Snapshot 7 (DataVersion 5009) block palette (string ids and `id`/`properties`), chunk status `minecraft:terrain`, and empty sign JSON. Older saves still read and write in their original format. Recent worlds can be cleared all at once, or removed one by one with the top-right cross.
- **Update**: Align all 16 locales with `en.lang` (including Traditional Chinese Options: FOV, render distance, camera sensitivity). Paste can align from the centre or a corner. Recent worlds can be sorted by last edited, created time, or version, in ascending or descending order.


gentlegiantJGC:
I suspect this is AI generated. I won't accept AI generated translations.
Moving hard coded strings into the language files is a good idea but this would be better split into at least a dozen smaller pull requests.
There are also a load of other changes that have nothing to do with internationalisation.

